PR #5, opened 2019-03-22
Make bound IP address configurable

nbaksalyar:


src/lib.rs
```diff
@@ -86,6 +86,7 @@ impl Crust {
     /// very limited functionaity will be available.
     pub fn start_listening(&mut self) {
         let port = self.cfg.port.unwrap_or(0);
+        let ip = self.cfg.ip.unwrap_or_else(|| Ipv4Addr::new(127, 0, 0, 1));
```
ustulation: This should be "0.0.0.0" by default

src/lib.rs
```diff
@@ -86,6 +86,7 @@ impl Crust {
     /// very limited functionaity will be available.
     pub fn start_listening(&mut self) {
         let port = self.cfg.port.unwrap_or(0);
+        let ip = self.cfg.ip.unwrap_or_else(|| Ipv4Addr::new(0, 0, 0, 0));
```
douglascaetano: You can use `unwrap_or()` instead.

```suggestion
        let ip = self.cfg.ip.unwrap_or(Ipv4Addr::new(0, 0, 0, 0));
```

nbaksalyar: I believe `cargo clippy` would complain in this case:
```
warning: use of `unwrap_or` followed by a function call
  --> src/lib.rs:89:30
   |
89 |         let ip = self.cfg.ip.unwrap_or(Ipv4Addr::new(0, 0, 0, 0));
   |                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `unwrap_or_else(|| Ipv4Addr::new(0, 0, 0, 0))`
```

ustulation: Yes unwrap_or will always create the default for you even if you don't need it. So it's a real waste if the construction was huge/expensive. In this case though that is debatable but i guess clippy does not do deeper analysis on the type

PR #7, opened 2019-03-22
Add travis CI script

povilasb:
It checks formatting and run tests, but omits the clippy checks because there are many warnings that have to be fixed.
Will do this in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,22 @@
+env:
+  global:
+    - RUST_BACKTRACE=1
+    - PATH=$PATH:$HOME/.cargo/bin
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,22 @@
...
+os:
+  - linux
+  - osx
+  - windows
+language: rust
+sudo: false
+cache:
+  cargo: true
+before_script:
+  - if ! cargo prune --version | grep -q "0.1.8"; then
```
ustulation: can you pls use the latest one - that has some improvements i did with JP

PR #9, opened 2019-03-28
Update quinn

povilasb:
Latest quinn has some fixes but also changes semantics on how
connections should be handled.  Now each incoming/outgoing connection
has its own driver which must be executed in order for connection to
proceed.

### DIFF
--- a/src/communicate.rs
+++ b/src/communicate.rs
@@ -306,6 +306,7 @@ fn handle_echo_resp(our_ext_addr: SocketAddr, inform_tx: Option<Sender<SocketAdd
     }
 }
 
+/// When connection fails, remove it to prevent memory leaks.
 fn handle_communication_err(peer_addr: SocketAddr, e: &Error, details: &str) {
     println!(
         "ERROR in communication with peer {}: {:?} - {}. Details: {}",
--- a/src/communicate.rs
+++ b/src/communicate.rs
@@ -312,11 +312,5 @@ fn handle_communication_err(peer_addr: SocketAddr, e: &Error, details: &str) {
         "ERROR in communication with peer {}: {:?} - {}. Details: {}",
         peer_addr, e, e, details
     );
-    ctx_mut(|c| {
-        if c.connections.remove(&peer_addr).is_some() {
-            if let Err(e) = c.event_tx.send(Event::ConnectionFailure { peer_addr }) {
-                println!("Could not fire event to the user: {}", e);
-            }
-        }
-    })
+    let _ = ctx_mut(|c| c.connections.remove(&peer_addr));
 }
--- a/src/connect.rs
+++ b/src/connect.rs
@@ -72,15 +72,21 @@ pub fn connect_to(peer_info: CrustInfo, send_after_connect: Option<WireMsg>) -> 
 
 fn handle_new_connection_res(
     peer_addr: SocketAddr,
-    new_peer_conn_res: Result<quinn::NewClientConnection, quinn::ConnectionError>,
+    new_peer_conn_res: Result<quinn::NewConnection, quinn::ConnectionError>,
 ) {
     let new_peer_conn = match new_peer_conn_res {
         Ok(new_peer_conn) => new_peer_conn,
         Err(e) => return handle_connect_err(peer_addr, &From::from(e)),
     };
+    current_thread::spawn(
+        new_peer_conn
+            .driver
+            .map_err(move |e| handle_connect_err(peer_addr, &From::from(e))),
+    );
 
     println!("Successfully connected to peer: {}", peer_addr);
 
+    let q_conn = new_peer_conn.connection;
     ctx_mut(|c| {
         let conn = match c.connections.get_mut(&peer_addr) {
             Some(conn) => conn,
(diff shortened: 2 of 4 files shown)


src/connect.rs
```diff
@@ -72,15 +72,17 @@ pub fn connect_to(peer_info: CrustInfo, send_after_connect: Option<WireMsg>) ->
...
 ) {
     let new_peer_conn = match new_peer_conn_res {
         Ok(new_peer_conn) => new_peer_conn,
         Err(e) => return handle_connect_err(peer_addr, &From::from(e)),
     };
+    current_thread::spawn(new_peer_conn.driver.map_err(|_| ()));
 
     println!("Successfully connected to peer: {}", peer_addr);
 
+    let q_conn = new_peer_conn.connection;
```
ustulation: Why is this diff necessary ?

src/connect.rs
```diff
@@ -72,15 +72,17 @@ pub fn connect_to(peer_info: CrustInfo, send_after_connect: Option<WireMsg>) ->
...
 fn handle_new_connection_res(
     peer_addr: SocketAddr,
-    new_peer_conn_res: Result<quinn::NewClientConnection, quinn::ConnectionError>,
+    new_peer_conn_res: Result<quinn::NewConnection, quinn::ConnectionError>,
 ) {
     let new_peer_conn = match new_peer_conn_res {
         Ok(new_peer_conn) => new_peer_conn,
         Err(e) => return handle_connect_err(peer_addr, &From::from(e)),
     };
+    current_thread::spawn(new_peer_conn.driver.map_err(|_| ()));
```
ustulation: In case of error I would like to destroy the connection object to prevent leaks and limbo states. I don't know what this error means but i guess dropping it is the safest way to go ?

src/listener.rs
```diff
@@ -22,6 +22,7 @@ fn handle_new_conn(new_conn: quinn::NewConnection) {
     // send in their certificate - have a delay and then log and forget them
     let q_conn = new_conn.connection;
     let incoming_streams = new_conn.incoming;
+    current_thread::spawn(new_conn.driver.map_err(|_| ()));
```
ustulation: Same as with the `connect.rs` review above:
> In case of error I would like to destroy the connection object to prevent leaks and limbo states. I don't know what this error means but i guess dropping it is the safest way to go ?

src/connect.rs
povilasb: Otherwise `new_peer_conn` would be used inside `ctx_mut({})` context which is impossible, because `current_thread::spawn(new_peer_conn.driver.map_err(|_| ()));` moves `driver` from `new_peer_conn`

src/connect.rs
```diff
@@ -72,15 +72,19 @@ pub fn connect_to(peer_info: CrustInfo, send_after_connect: Option<WireMsg>) ->
...
     peer_addr: SocketAddr,
-    new_peer_conn_res: Result<quinn::NewClientConnection, quinn::ConnectionError>,
+    new_peer_conn_res: Result<quinn::NewConnection, quinn::ConnectionError>,
 ) {
     let new_peer_conn = match new_peer_conn_res {
         Ok(new_peer_conn) => new_peer_conn,
         Err(e) => return handle_connect_err(peer_addr, &From::from(e)),
     };
+    current_thread::spawn(new_peer_conn.driver.map_err(move |e| {
+        communicate::handle_communication_err(peer_addr, &From::from(e), "Connection driver failed")
```
ustulation: this should call `handle_connect_err` as that is specialised to deal with connect errors

src/listener.rs
```diff
@@ -22,9 +22,12 @@ fn handle_new_conn(new_conn: quinn::NewConnection) {
     // send in their certificate - have a delay and then log and forget them
     let q_conn = new_conn.connection;
     let incoming_streams = new_conn.incoming;
-
     let peer_addr = q_conn.remote_address();
 
+    current_thread::spawn(new_conn.driver.map_err(move |e| {
+        communicate::handle_communication_err(peer_addr, &From::from(e), "Connection driver failed")
```
ustulation: This should do what this is doing:
https://github.com/ustulation/using-quinn/blob/df0d633103f579c5b1bcb84dd600d345c35a52f8/src/listener.rs#L88
not call anything from the `communicate` module as that also fires an event which we might not need (if someone was incoming and got disconnected before doing anything, user didn't even know about it in the first place so no use firing a `ConnectionFailure` event to them)

You can create an extra function here to direct both of these (the one above and the one linked) to that if you want to print it but do keep it confined to the module (in this case `listener.rs`). Then later it makes it easier to analyse if we can combine something.

src/communicate.rs
```diff
@@ -306,7 +306,8 @@ fn handle_echo_resp(our_ext_addr: SocketAddr, inform_tx: Option<Sender<SocketAdd
     }
 }
 
-fn handle_communication_err(peer_addr: SocketAddr, e: &Error, details: &str) {
+/// When connection fails, remove it to prevent memory leaks.
+pub fn handle_communication_err(peer_addr: SocketAddr, e: &Error, details: &str) {
```
ustulation: Doesn't need to be public - see other reviews.
While at it can you pls change:
https://github.com/ustulation/using-quinn/blob/master/src/communicate.rs#L315-L319
to simply remove only (like said for the review in `listener.rs`) as otherwise it's going to cause a spurious fire.
So only `let _ = c.connections.remove(..);`

PR #10, opened 2019-04-01
Update to latest quinn

povilasb:
* This quinn version uses miliseconds rather than seconds for time based values;
* it has fixed packet timeout handling which used to stall message
  passing;
* Connection::close() now consumes connection, but there's no need to
  explicitly close the connection as that will happen when `Connection`
  object is dropped.

### DIFF
--- a/src/connect.rs
+++ b/src/connect.rs
@@ -72,21 +72,25 @@ pub fn connect_to(peer_info: CrustInfo, send_after_connect: Option<WireMsg>) -> 
 
 fn handle_new_connection_res(
     peer_addr: SocketAddr,
-    new_peer_conn_res: Result<quinn::NewConnection, quinn::ConnectionError>,
+    new_peer_conn_res: Result<
+        (
+            quinn::ConnectionDriver,
+            quinn::Connection,
+            quinn::IncomingStreams,
+        ),
+        quinn::ConnectionError,
+    >,
 ) {
-    let new_peer_conn = match new_peer_conn_res {
-        Ok(new_peer_conn) => new_peer_conn,
+    let (conn_driver, q_conn) = match new_peer_conn_res {
+        Ok((conn_driver, q_conn, _)) => (conn_driver, q_conn),
         Err(e) => return handle_connect_err(peer_addr, &From::from(e)),
     };
     current_thread::spawn(
-        new_peer_conn
-            .driver
-            .map_err(move |e| handle_connect_err(peer_addr, &From::from(e))),
+        conn_driver.map_err(move |e| handle_connect_err(peer_addr, &From::from(e))),
     );
 
     println!("Successfully connected to peer: {}", peer_addr);
 
-    let q_conn = new_peer_conn.connection;
     ctx_mut(|c| {
         let conn = match c.connections.get_mut(&peer_addr) {
             Some(conn) => conn,
--- a/src/connect.rs
+++ b/src/connect.rs
@@ -25,8 +25,8 @@ pub fn connect_to(peer_info: CrustInfo, send_after_connect: Option<WireMsg>) -> 
         let transport_config = unwrap!(Arc::get_mut(&mut peer_cfg.transport));
         // TODO test that this is sent only over the uni-stream to the peer not on the uni
         // stream from the peer
-        transport_config.idle_timeout = ctx(|c| c.idle_timeout);
-        transport_config.keep_alive_interval = ctx(|c| c.keep_alive_interval);
+        transport_config.idle_timeout = ctx(|c| c.idle_timeout) * 1000;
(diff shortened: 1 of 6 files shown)


src/connect.rs
```diff
@@ -25,8 +25,8 @@ pub fn connect_to(peer_info: CrustInfo, send_after_connect: Option<WireMsg>) ->
         let transport_config = unwrap!(Arc::get_mut(&mut peer_cfg.transport));
         // TODO test that this is sent only over the uni-stream to the peer not on the uni
         // stream from the peer
-        transport_config.idle_timeout = ctx(|c| c.idle_timeout);
-        transport_config.keep_alive_interval = ctx(|c| c.keep_alive_interval);
+        transport_config.idle_timeout = ctx(|c| c.idle_timeout) * 1000;
```
ustulation: If they have changed to msec can we change the context struct and config struct to reflect this (so let that number denote msec instead of secs like now, and consequently not multiply by 1000 here)

0)
Chnage https://github.com/ustulation/using-quinn/blob/48c9799c69fc28f139ff0cb2e74bb17ae058011d/src/context.rs#L76 to
```
        idle_timeout_msec: u64,
```
1)
https://github.com/ustulation/using-quinn/blob/48c9799c69fc28f139ff0cb2e74bb17ae058011d/src/context.rs#L87 to
```
        keep_alive_interval_msec: u32,
```
2)
https://github.com/ustulation/using-quinn/blob/48c9799c69fc28f139ff0cb2e74bb17ae058011d/src/config.rs#L21 to
```
        pub idle_timeout_msec: Option<u64>,
```
and https://github.com/ustulation/using-quinn/blob/48c9799c69fc28f139ff0cb2e74bb17ae058011d/src/config.rs#L20 to
```
        /// The interval is in miliseconds. A value of 0 disables this feature.
```
3) same for `keep_alive_interval` and its doc-comments

src/context.rs
```diff
@@ -183,6 +183,14 @@ impl ToPeer {
     }
 }
 
+impl Drop for ToPeer {
```
ustulation: Why re-arranging this - dtor usually comes last in the traits applied ? Anyway this is producing unnecessary diff

src/context.rs
```diff
@@ -244,6 +244,14 @@ impl FromPeer {
     }
 }
 
+impl Drop for FromPeer {
```
ustulation: This is producing unnecessary diff. Only clone() needed (which too i think it might be no longer as upstream has agreed to not consume it)

src/lib.rs
```diff
@@ -124,13 +124,13 @@ impl Crust {
                 let transport_config = unwrap!(Arc::get_mut(&mut our_cfg.transport_config));
                 // TODO test that this is sent only over the uni-stream to the peer not on the uni
                 // stream from the peer
-                transport_config.idle_timeout = idle_timeout;
-                transport_config.keep_alive_interval = keep_alive_interval;
+                transport_config.idle_timeout = idle_timeout * 1000;
```
ustulation: same as `connect.rs` - not needed after the changes mentioned

src/connect.rs
```diff
@@ -72,21 +72,25 @@ pub fn connect_to(peer_info: CrustInfo, send_after_connect: Option<WireMsg>) ->
...
+            quinn::Connection,
+            quinn::IncomingStreams,
+        ),
+        quinn::ConnectionError,
+    >,
 ) {
-    let new_peer_conn = match new_peer_conn_res {
-        Ok(new_peer_conn) => new_peer_conn,
+    let (conn_driver, q_conn) = match new_peer_conn_res {
+        Ok((conn_driver, conn, _)) => (conn_driver, conn),
```
ustulation: I was reserving `conn` for our `Connection` struct, `new_peer_conn` for `quinn::NewConnection` and `q_conn` for `quinn::Connection`. So let's stick to that pls and retain the name of `new_conn` here.

src/listener.rs
```diff
@@ -11,22 +11,24 @@ use tokio::runtime::current_thread;
 pub fn listen(incoming_connections: quinn::Incoming) {
     let leaf = incoming_connections
         .map_err(|()| println!("ERROR: Listener errored out"))
-        .for_each(move |new_conn| {
-            handle_new_conn(new_conn);
+        .for_each(move |(conn_driver, conn, incoming)| {
```
ustulation: Let's call this `new_peer_conn` for `quinn::NewConnection` and `q_conn` for `quinn::Connection`. `conn` is for our `Connection`: https://github.com/ustulation/using-quinn/blob/48c9799c69fc28f139ff0cb2e74bb17ae058011d/src/listener.rs#L39

PR #12, opened 2019-04-09
Update chat example

povilasb:
Added "sendrand" command that helps testing QUIC multistreaming in Crust. This command allows to send random messages with the given size.

### DIFF
--- a/examples/chat.rs
+++ b/examples/chat.rs
@@ -13,12 +13,13 @@
 extern crate unwrap;
 
 use std::net::IpAddr;
-use std::sync::mpsc::channel;
+use std::sync::mpsc::{channel, Receiver};
 use std::sync::{Arc, Mutex};
-use std::thread;
+use std::thread::{self, JoinHandle};
 
 use bytes::Bytes;
 use clap::{App, Arg};
+use rand::{self, RngCore};
 use rustyline::config::Configurer;
 use rustyline::error::ReadlineError;
 use rustyline::Editor;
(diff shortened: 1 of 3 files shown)


examples/chat.rs
```diff
@@ -172,3 +206,63 @@ fn print_ourinfo(crust: &mut Crust) {
...
+    let port = matches
+        .value_of("listening_port")
+        .and_then(|v| v.parse().ok());
+    let our_ip = matches.value_of("our_ip").map(|addr| unwrap!(addr.parse()));
+
+    CliArgs { port, our_ip }
+}
+
+fn print_logo() {
+    println!(
```
ustulation: `quic-chat` as this will be a separate crate to crust

examples-bkup/chat.rs
```diff
@@ -172,3 +207,64 @@ fn print_ourinfo(crust: &mut Crust) {
         serde_json::to_string(&ourinfo).unwrap()
     );
 }
+
+fn parse_cli_args() -> CliArgs {
+    let matches = App::new("Simple chat app built on Crust")
```
nbaksalyar: I think this should be
```suggestion
    let matches = App::new("Simple chat app built on p2p-quic")
```

examples-bkup/chat.rs
```diff
@@ -172,3 +207,64 @@ fn print_ourinfo(crust: &mut Crust) {
...
+        .about(
+            "This chat app connects two machines directly without intermediate servers and allows \
+             to exchange messages securely. All the messages are end to end encrypted.",
+        )
+        .arg(
+            Arg::with_name("listening_port")
+                .help(
+                    "Optional server Crust will be listening for incoming connections ON. If \
```
nbaksalyar: Same thing about Crust. And it probably could use a bit of a rephrasing:
```suggestion
                    "Optional port number which the peer will use to listen for incoming connections. If \
```

examples-bkup/chat.rs
```diff
@@ -172,3 +207,64 @@ fn print_ourinfo(crust: &mut Crust) {
...
+                .short("p")
+                .value_name("PORT")
+                .takes_value(true)
+                .validator(|v| v.parse::<u16>().and(Ok(())).or(Err("Invalid port".into()))),
+        )
+        .arg(
+            Arg::with_name("our_ip")
+                .help(
+                    "Our IP address to use when constructing our connection info. If unspecified \
+                     , Crust will try to use bootstrap node to determine our IP.",
```
nbaksalyar: Probably `Crust` can be replaced with just `peer`
```suggestion
                     , peer will try to use bootstrap node to determine our IP.",
```

PR #13, opened 2019-04-09
Fix examples and add deployment scripts

nbaksalyar:
Also renames `Crust` and `CrustInfo` to `Peer` and `PeerInfo`


povilasb:
Deprecated in favor of https://github.com/maidsafe/quic-p2p/pull/19 and https://github.com/maidsafe/quic-p2p/pull/16

PR #14, opened 2019-04-10
Bootstrap cache

povilasb:


src/dirs.rs
```diff
@@ -0,0 +1,57 @@
...
+    pub fn new(path: &str) -> Self {
+        Self {
+            path: PathBuf::from(path.to_owned()),
+        }
+    }
+}
+
+pub enum Dirs {
+    #[cfg(not(any(target_os = "android", target_os = "androideabi", target_os = "ios")))]
+    Desktop(ProjectDirs),
```
ustulation: Can we match the enum order to the order in the functions ? It seems all functions put `Desktop` as second - we should perhaps have it here in that order too - usually makes looking at stuff easier (and i suspect most of our code is thus)

src/lib.rs
```diff
@@ -211,7 +225,14 @@ impl Crust {
             if our_type != OurType::Client {
                 listener::listen(incoming_connections);
             }
+
+            let _ = err_tx.send(Ok(()));
         });
+
+        err_rx
+            .recv()
+            .map_err(Error::ChannelRecv)
```
ustulation: If you already have a `from()` conversion what's the use of mapping the error - why not simply `Ok(blah()?)` kind of semantics ?

src/bootstrap_cache.rs
```diff
@@ -0,0 +1,163 @@
...
+use log::{error, info};
+use std::fs::{self, File};
+use std::io::{self, BufReader, BufWriter};
+use std::path::Path;
+use std::path::PathBuf;
+
+/// Maximum peers in the cache.
+const MAX_CACHE_SIZE: usize = 200;
+
+type Stack = Vec<NodeInfo>;
```
ustulation: I wouldn't typdef this as it's used in just this file so seems like an overkill. But the main reason is that it's not really a stack - there's no LIFO semantics - in-fact quite the opposite with shuffling and all and also we are least likely to remove someone we added last as that's the freshest entry.

src/bootstrap_cache.rs
```diff
@@ -0,0 +1,163 @@
...
+        } else {
+            self.insert_new(peer);
+        }
+    }
+
+    fn insert_new(&mut self, peer: NodeInfo) {
+        self.stack.push(peer);
+        self.add_count += 1;
+        if self.stack.len() == MAX_CACHE_SIZE {
+            let _ = self.stack.pop();
```
ustulation: You are popping someone you just added - that's wrong.

src/bootstrap_cache.rs
```diff
@@ -0,0 +1,163 @@
...
+        }
+        self.try_sync_to_disk();
+    }
+
+    fn move_to_cache_top(&mut self, peer: NodeInfo) {
+        // remove and push back
+        let mut t: Stack = self.stack.iter().filter(|&x| *x == peer).cloned().collect();
+        t.push(peer);
+        self.stack = t;
```
ustulation: Seems a very wasteful way to achieve the goal of just putting someone to the top. Why not just find and remove from that index and push it to the same vec instead of a fresh new allocation etc.

So
```
if let Some(info) = stack.remove(n) { stack.push(info); }
// n can be obtained via `pos()` or check if you can find and remove at the same time
```

Goes without saying if `n == len - 1` then you don't really have to do anything.

src/listener.rs
```diff
@@ -61,6 +61,7 @@ fn handle_new_conn(
                     peer_cert_der: peer_cert_der.clone(),
                 };
 
+                c.bootstrap_cache.add_peer(node_info.clone());
```
ustulation: Maybe at-least allow some message exchanges (so only on read) before adding to a cache instead of adding anyone who connects to us (and maybe just to displace valid people from our bootstrap cache). This makes it even easier if done on `listen()`.

src/communicate.rs
```diff
@@ -212,20 +223,30 @@ pub fn handle_wire_msg(peer_addr: SocketAddr, wire_msg: WireMsg) {
...
-                            wire_msg,
-                        ),
-                        ToPeer::Established { ref q_conn, .. } => dispatch_wire_msg(
-                            peer_addr,
+                            Peer::Client { peer_addr },
                             q_conn,
                             c.our_ext_addr_tx.take(),
                             &c.event_tx,
                             wire_msg,
+                            &mut c.bootstrap_cache,
```
ustulation: This is wrong - if `ToPeer` is not-needed then that peer is basically a client. Why would we want to add that to the bootstrap cache ?

src/bootstrap_cache.rs
```diff
@@ -0,0 +1,163 @@
...
+use std::path::Path;
+use std::path::PathBuf;
+
+/// Maximum peers in the cache.
+const MAX_CACHE_SIZE: usize = 200;
+
+type Stack = Vec<NodeInfo>;
+
+/// A very simple LRU like struct that writes itself to disk every 10 entries added.
+pub struct BootstrapCache {
```
ustulation: We should perhaps pass the hard-coded-contacts list to this to be held as a member-var. That way it can always see if the peer being asked to add belongs to HCC. If yes then ignore.

src/lib.rs
povilasb: In this case `from()` simply does not get applied to the error - don't know why though.

src/communicate.rs
povilasb: It won't be added to the cache - the code is just there to confirm with the `dispatch_wire_msg()` signature.
```
                        ToPeer::NotNeeded => dispatch_wire_msg(
                            Peer::Client { peer_addr },
                            q_conn,
                            c.our_ext_addr_tx.take(),
                            &c.event_tx,
                            wire_msg,
                            &mut c.bootstrap_cache,
                        ),
```
See we are telling `dispatch_wire_msg` that this is the `Client` peer and it will act accordingly.

src/bootstrap_cache.rs
```diff
@@ -0,0 +1,217 @@
...
+            self.move_to_cache_top(peer);
+        } else {
+            self.insert_new(peer);
+        }
+    }
+
+    fn insert_new(&mut self, peer: NodeInfo) {
+        self.peers.push_back(peer);
+        self.add_count += 1;
+        if self.peers.len() == MAX_CACHE_SIZE {
```
ustulation:
```suggestion
        if self.peers.len() > MAX_CACHE_SIZE {
```

src/connect.rs
```diff
@@ -164,23 +165,19 @@ fn handle_new_connection_res(
...
-
-                if let Err(e) = c.event_tx.send(Event::ConnectedTo { peer }) {
+                if let Err(e) = c.event_tx.send(Event::ConnectedTo { peer: peer.clone() }) {
                     println!("Could not fire event: {:?}", e);
                 }
 
                 for pending_read in pending_reads.drain(..) {
                     communicate::dispatch_wire_msg(
-                        peer_addr,
+                        peer.clone(),
```
ustulation:
```suggestion
                        peer,
```

src/lib.rs
```diff
@@ -323,6 +346,18 @@ impl QuicP2p {
     }
 }
 
+/// Tries to determine the best location for bootstrap cache and constructs it.
+fn init_bootstrap_cache(mut hard_coded_contacts: Vec<NodeInfo>) -> R<BootstrapCache> {
```
ustulation: Is this more appropriate in the bootstrap cache module ?

src/communicate.rs
```diff
@@ -196,13 +197,23 @@ pub fn handle_wire_msg(peer_addr: SocketAddr, wire_msg: WireMsg) {
...
-                        ),
+                        ToPeer::Established {
+                            ref q_conn,
+                            ref peer_cert_der,
+                        } => {
+                            let node_info = NodeInfo {
+                                peer_addr,
+                                peer_cert_der: peer_cert_der.clone(),
+                            };
+                            dispatch_wire_msg(
```
ustulation: As was discussed this can be exploited and good peers displaced from the cache by anyone. Do not update bootstrap cache here. Do only when upper layer tell us to contact them (e.g, write something to them and we have successfully connected to them _because the user layer wanted us to_ (not because of automatic reverse connection) and after that every time the _user sends something_ (so not automatic responses to the incoming messages such as echo responses etc). So never on read.

src/connect.rs
povilasb: Can't do that: `dispatch_wire_msg()` is called multiple times in a loop and `Peer` does not implement `Copy`.

src/context.rs
```diff
@@ -119,6 +123,14 @@ impl Context {
 pub struct Connection {
     pub to_peer: ToPeer,
     pub from_peer: FromPeer,
+
+    /// quic-p2p won't validate incoming peers, it will simply pass them to the upper layer.
+    /// Until we know that these peers are useful/valid for the upper layers, we might refrain
+    /// ourselves from taking specific actions: e.g. putting these peers into the bootstrap cache.
+    /// This flag indicates whether upper layer attempted to connect/send something to the other
+    /// end of this connection.
+    pub we_contacted_peer: bool,
```
ustulation: Where are we actually setting this ?

PR #26, opened 2019-04-16
Add more tests

povilasb:
Adds some bootstrap cache related tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,10 @@ before_script:
     fi
   - rustup component add rustfmt clippy
 script:
-  - cargo fmt -- --check &&
-    cargo test --verbose --release
+  - cargo fmt -- --check
+  - if [ "${TRAVIS_OS_NAME}" = linux ]; then
+      cargo clippy --verbose --release --all-targets ;
+    fi
+  - cargo test --verbose --release
 before_cache:
   - cargo prune
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -41,3 +41,36 @@ pub fn handle_communication_err(peer_addr: SocketAddr, e: &Error, details: &str)
     );
     let _ = ctx_mut(|c| c.connections.remove(&peer_addr));
 }
+
+#[cfg(test)]
+pub mod testing {
+    use crate::config::SerialisableCertificate;
+    use crate::dirs::{Dirs, OverRide};
+    use crate::NodeInfo;
+    use rand::Rng;
+    use std::env;
+    use std::net::{Ipv4Addr, SocketAddr, SocketAddrV4};
+    use std::path::PathBuf;
+
+    pub fn test_dirs() -> Dirs {
+        Dirs::Overide(OverRide::new(&unwrap!(tmp_dir().to_str())))
+    }
+
+    pub fn rand_node_info() -> NodeInfo {
+        let peer_cert_der = SerialisableCertificate::default().cert_der;
+        let mut rng = rand::thread_rng();
+        let port: u16 = rng.gen();
+        let peer_addr = SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::new(127, 0, 0, 1), port));
+        NodeInfo {
+            peer_addr,
+            peer_cert_der,
+        }
+    }
+
+    fn tmp_dir() -> PathBuf {
+        let fname = format!("{:016x}.quic_p2p_tests", rand::random::<u64>());
+        let mut path = env::temp_dir();
+        path.push(fname);
+        path
+    }
+}
--- a/tests/quic_p2p.rs
+++ b/tests/quic_p2p.rs
@@ -1,0 +1,59 @@
+use quic_p2p::{Builder, Config, Event, Peer, QuicP2p};
+use std::net::{IpAddr, Ipv4Addr};
+use std::sync::mpsc;
+use unwrap::unwrap;
+
+/// Waits for `Event::ConnectedTo`.
+fn wait_till_connected(ev_rx: mpsc::Receiver<Event>) -> Peer {
+    for event in ev_rx.iter() {
+        if let Event::ConnectedTo { peer } = event {
+            return peer;
+        }
+    }
+    panic!("Didn't receive the expected ConnectodTo event");
+}
+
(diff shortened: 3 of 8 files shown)


tests/quic_p2p.rs
```diff
@@ -0,0 +1,40 @@
+use quic_p2p::{Config, Event, QuicP2p};
+use std::net::{IpAddr, Ipv4Addr};
+use std::sync::mpsc;
+use unwrap::unwrap;
+
+#[test]
+fn successfull_connection_stores_peer_in_bootstrap_cache() {
+    let (ev_tx, _ev_rx) = mpsc::channel();
```
ustulation: Maybe use local scope instead of variable shadowing (`ev_tx` is shadowed later):
```
let ... = {
  // some code
  QuicP2p::with_config(...)
};
```

It keeps surprises minimum if someone changed code.

tests/quic_p2p.rs
```diff
@@ -0,0 +1,40 @@
...
+    peer2.connect_to(peer1_conn_info.clone());
+
+    for event in ev_rx.iter() {
+        if let Event::ConnectedTo { .. } = event {
+            break;
+        }
+    }
+
+    let cache = unwrap!(peer2.bootstrap_cache());
+    assert_eq!(cache, vec![peer1_conn_info]);
```
ustulation: Maybe also assert that `peer1` OTOH did also get `ConnectedTo` but still did not add `peer2` in its cache.

tests/quic_p2p.rs
```diff
@@ -0,0 +1,40 @@
...
+            port: Some(0),
+            ip: Some(IpAddr::V4(Ipv4Addr::LOCALHOST)),
+            ..Default::default()
+        },
+    );
+    unwrap!(peer2.start_listening());
+    peer2.connect_to(peer1_conn_info.clone());
+
+    for event in ev_rx.iter() {
+        if let Event::ConnectedTo { .. } = event {
```
ustulation: Good to assert that peers indeed connected to the expected other peer (so make use of the info in the event)

tests/quic_p2p.rs
povilasb: I'll make this a separate test case - don't want a single test check too many things :)

ustulation: Hmm it's just two lines i guess. Not sure about the verbosity that so many test cases add. By similar logic routing probably would have 1000s of test cases. I think a single test case is far easier to look at as the reader can maintain a flow while evaluating/reading it. Having too many small test cases just to prove what could be done is a few extra lines in the same test case seems overkill to me. I don't know what the best practice is though.

src/utils.rs
```diff
@@ -41,3 +41,36 @@ pub fn handle_communication_err(peer_addr: SocketAddr, e: &Error, details: &str)
...
+    use rand::Rng;
+    use std::env;
+    use std::net::{Ipv4Addr, SocketAddr, SocketAddrV4};
+    use std::path::PathBuf;
+
+    pub fn test_dirs() -> Dirs {
+        Dirs::Overide(OverRide::new(&unwrap!(tmp_dir().to_str())))
+    }
+
+    pub fn rand_peer() -> NodeInfo {
```
ustulation: In integration-test in this PR a test-peer means a fully constructed `QuicP2p` instance ready to be used. This name is in conflict with that - so better to call this `rand_node_info()`

Also not calling it peer (in the light of what test-peer above does) makes sense as it's not usable really as the random port is chosen by you not the OS (not problems like getting in a range where permission is denied or already occupied etc,) so might not be useful for creating an actual peer.

tests/quic_p2p.rs
ustulation: Hmm https://github.com/maidsafe/quic-p2p/pull/26/files#r275857033 is still not covered with in this update of the PR. So pls confirm that point (sp. the 2nd part of the sentence which i'm mainly interested in)

src/utils.rs
```diff
@@ -41,3 +41,36 @@ pub fn handle_communication_err(peer_addr: SocketAddr, e: &Error, details: &str)
...
+        let peer_cert_der =
+            rcgen::generate_simple_self_signed(vec!["Test".to_string()]).serialize_der();
```
ustulation: Can't we use `Config::with_default_cert()` and not use rc-gens and the rest as that does all that internally, unless the intention was to create a special Certificate - which doens't seem so here as it's just a random one.

.travis.yml
```diff
@@ -18,5 +18,8 @@ before_script:
 script:
   - cargo fmt -- --check &&
     cargo test --verbose --release
+  - if [ "${TRAVIS_OS_NAME}" = linux ]; then
```
ustulation: Seems wasteful (slower) to do it here ? Better to put it after `cargo fmt` check and before `cargo test`. It's more like head of the queue blocking where something that could complete much faster is delayed by something which is much slower to complete.

src/utils.rs
povilasb: How about `SerialisableCertificate::default()` since I don't need anything else from the config.

tests/quic_p2p.rs
povilasb: Good catch, there was actually a bug, fixed it :)

src/utils.rs
```diff
@@ -41,3 +41,36 @@ pub fn handle_communication_err(peer_addr: SocketAddr, e: &Error, details: &str)
...
+
+    pub fn test_dirs() -> Dirs {
+        Dirs::Overide(OverRide::new(&unwrap!(tmp_dir().to_str())))
+    }
+
+    pub fn rand_node_info() -> NodeInfo {
+        let peer_cert_der = SerialisableCertificate::default().cert_der;
+        let mut rng = rand::thread_rng();
+        let port: u16 = rng.gen();
+        let peer_addr = SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::new(127, 0, 0, 1), port));
```
ustulation: Just use the constant `Ipv4Addr::LOCALHOST` like in production

ustulation: (No worries i'll do it in the PR i'm going to raise soon)

PR #34, opened 2019-04-18
Resolves #24 - Bootstrap and misc

### LINKED ISSUE #24: Bootstrap API
Provide an API to bootstrap. This will first use the bootstrap cache followed by hard-coded contacts to bootstrap off a proxy. Once the bootstrap with a proxy is successful we should kill all the ongoing bootstrap attempts to other proxies.

ustulation:
See individual commit

### DIFF
--- a/src/bootstrap.rs
+++ b/src/bootstrap.rs
@@ -1,0 +1,31 @@
+// Copyright 2019 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under the MIT license <LICENSE-MIT
+// http://opensource.org/licenses/MIT> or the Modified BSD license <LICENSE-BSD
+// https://opensource.org/licenses/BSD-3-Clause>, at your option. This file may not be copied,
+// modified, or distributed except according to those terms. Please review the Licences for the
+// specific language governing permissions and limitations relating to use of the SAFE Network
+// Software.
+
+use crate::connect;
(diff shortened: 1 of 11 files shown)


src/bootstrap.rs
```diff
@@ -0,0 +1,31 @@
...
+use crate::context::{ctx, BootstrapGroupBuilder};
+
+pub fn initiate() {
+    let (proxies, event_tx): (Vec<_>, _) = ctx(|c| {
+        (
+            c.bootstrap_cache
+                .peers()
+                .iter()
+                .rev()
+                .chain(c.bootstrap_cache.hard_coded_contacts().iter())
```
povilasb: Note that this will attempt to connect to hard coded contacts in parallel with the cached nodes. In case of success, I believe we'd like to postpone connection to hard coded contacts when connection to all cached nodes fails.

ustulation:
Ya i had thought about that - it seems OK to me .. we are going
sequentially first exhausting all the cache contacts before reaching the
hard coded ones .. so although all the connection attempts might be active
at the same time, they were sequentially spawned. Now if all the (200)
cached contacts were so bad that the Hard coded ones even though spawned
later in seq. beat them then so be it was my reasoning. While we are
prefering to start with cached contacts, introducing an artificial delay by
waiting for all of them to fail seems like deliberate crippling and slowing
down for the sake of it.

PR #36, opened 2019-04-19
Modularity

ustulation:


src/utils.rs
```diff
@@ -52,6 +66,30 @@ pub fn handle_communication_err(peer_addr: SocketAddr, e: &Error, details: &str)
     let _ = ctx_mut(|c| c.connections.remove(&peer_addr));
 }
 
+/// Try reading from the disk into the given structure.
+pub fn read_from_disk<D>(file_path: &Path) -> R<D>
+where
+    D: DeserializeOwned,
+{
+    Ok(File::open(file_path)
+        .map_err(|e| e.into())
```
nbaksalyar: Can be just `Into::into` (I think clippy should pick this up)

src/utils.rs
```diff
@@ -52,6 +66,30 @@ pub fn handle_communication_err(peer_addr: SocketAddr, e: &Error, details: &str)
...
+        .and_then(|mut rdr| bincode::deserialize_from(&mut rdr))?)
+}
+
+/// Try writing the given structure to the disk.
+pub fn write_to_disk<S>(file_path: &Path, s: &S) -> R<()>
+where
+    S: Serialize,
+{
+    File::create(file_path)
+        .map_err(|e| e.into())
```
nbaksalyar: Ditto for `Into::into`

src/wire_msg.rs
```diff
@@ -55,10 +55,10 @@ impl fmt::Display for WireMsg {
     }
 }
 
-/// Type of peer connecting to us.
+/// Type of Handshake.
 ///
-/// If the peer is a client then we allow a single connection between us with a bidirectional
-/// stream, otherwise we will have 2 connections with uni-directional stream in each.
+/// If the peer is a client then we allow a single connection between us, otherwise we will have 2
```
nbaksalyar: Maybe worth adding this extra bit of info here?
```suggestion
/// If the peer is a client then we allow a single connection between us which will open 2 unidirectional streams for each side, otherwise we will have 2
/// connections with only one uni-directional stream in each.
```
or something like this

PR #37, opened 2019-04-19
Test bootstrap

povilasb:
Add bootstrap() integration tests.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -193,7 +193,15 @@ impl fmt::Display for NodeInfo {
 }
 
 impl QuicP2p {
-    /// Bootstrap to a proxy
+    /// Bootstrap to the network.
+    ///
+    /// Bootstrap concept is interchangeable with "connect" except in the context quic-p2p
+    /// bootstrapping is more powerful than a simple `connect_to()`. `bootstrap()` will try to
+    /// connect to all peers that are specified in config (`hard_coded_contacts`) or were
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -193,7 +193,15 @@ impl fmt::Display for NodeInfo {
 }
 
 impl QuicP2p {
-    /// Bootstrap to a proxy
+    /// Bootstrap to the network.
+    ///
+    /// Bootstrap concept is interchangeable with "connect" except in the context quic-p2p
```
ustulation: It is not - it is different that a connected peer can only be disconnected either by remote going offline or explicit disconnect API. Bootstrap connects will all be auto disconnected except to one proxy even if the rest of the connections had succeeded in the meantime.

Maybe a good thing to have for tests:

0.  Above what the review says (so check that all but one proxy remains connected to at the end
1. Make sure that bootstrap cache peers are connected to first given both HCCs and BootstrapCache peers are present (so say have 5 BSCache and 5 HCC contacts and assert that one of the 1st 5 are always connected to first over local host)
2. Make all faulty peers that do not do handshake in the given amount of time. In that case bootstrap should fail _after_ Handshake timeout (as everyone was connected to but none proceeded for complete maturing of the connection)
3. Have clients and nodes both try and bootstrap (just now the tests test only node i think)

nbaksalyar:
Superseded by #41

PR #40, opened 2019-04-22
Use structopt for command line arguments

nbaksalyar:
Using this library gives us more flexibility in ways users can configure the options provided by the library.
This PR also includes some changes in `directories`/config file loader handling (e.g., it also takes mobile platforms into account).


src/utils.rs
```diff
@@ -13,7 +13,7 @@ use crate::error::Error;
 use serde::de::DeserializeOwned;
 use serde::Serialize;
 use std::fs::File;
-use std::io::{self, BufReader, BufWriter};
+use std::io::{BufReader, BufWriter};
```
ustulation: Why was this necessary

src/config.rs
```diff
@@ -14,34 +14,48 @@ use crate::{NodeInfo, R};
...
     /// The interval is in milliseconds. A value of 0 disables this feature.
+    #[structopt(long)]
     pub idle_timeout_msec: Option<u64>,
     /// Interval to send keep-alives if we are idling so that the peer does not disconnect from us
     /// declaring us offline. If none is supplied we'll default to the documented constant.
     ///
     /// The interval is in milliseconds. A value of 0 disables this feature.
+    #[structopt(long)]
     pub keep_alive_interval_msec: Option<u32>,
     /// Path to our TLS Certificate. This file must contain `SerialisableCertificate` as content
```
ustulation: Ah this is not correct - can we pls change it to say it's not a path we are asking but the certificate itself ?

src/utils.rs
nbaksalyar: It's not used on non-T1/mobile platforms (see the `project_dir` impl below)

PR #42, opened 2019-04-23
Add tests for malicious client scenarios

nbaksalyar:
1. Attempt to open a bi-directional QUIC stream;
2. Attempt to send an extra handshake message, reintroducing a client as a node and vice versa.

### DIFF
--- a/src/bootstrap.rs
+++ b/src/bootstrap.rs
@@ -30,3 +30,83 @@ pub fn start() {
         let _ = connect::connect_to(proxy, None, Some(&maker));
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::test_utils::new_random_qp2p_for_unit_test;
+    use crate::{Builder, Config, Event};
+    use std::collections::HashSet;
+    use std::net::{IpAddr, Ipv4Addr};
+    use std::sync::mpsc;
+    use std::thread;
+    use std::time::Duration;
+
+    // Try to bootstrap to 4 different peer sequentially. Use an artificial delay for 3 of the peers.
+    // Make sure in the end we have only one bootstrapped connection (all other connections are
+    // supposed to be dropped).
+    #[test]
+    fn bootstrap_to_only_one_node_with_delays() {
+        test_bootstrap_to_multiple_peers(4, Some(10));
+    }
+
+    // Try to bootstrap to 4 different peer at once.
+    // Make sure in the end we have only one bootstrapped connection (all other connections are
+    // supposed to be dropped).
+    #[test]
+    fn bootstrap_to_only_one_node_without_delays() {
+        test_bootstrap_to_multiple_peers(4, None);
+    }
+
+    fn test_bootstrap_to_multiple_peers(num_conns: u64, delay_ms: Option<u64>) {
+        let mut bs_nodes = Vec::with_capacity(num_conns as usize);
+        let mut bs_peer_addrs = HashSet::with_capacity(num_conns as usize);
+        let mut hcc_contacts = HashSet::with_capacity(num_conns as usize);
+
+        // Spin up 4 peers that we'll use for hardcoded contacts.
+        for i in 0..num_conns {
+            let (tx, rx) = mpsc::channel();
+            let mut qp2p = unwrap!(Builder::new(tx)
+                .with_config(Config {
+                    ip: Some(IpAddr::V4(Ipv4Addr::LOCALHOST)),
+                    port: Some(0),
+                    ..Default::default()
+                })
+                .build());
+
+            if let Some(delay_ms) = delay_ms {
+                qp2p.set_connect_delay(i * delay_ms);
+            }
+
+            let ci_info = unwrap!(qp2p.our_connection_info());
+
+            bs_peer_addrs.insert(ci_info.peer_addr);
+            hcc_contacts.insert(ci_info);
+            bs_nodes.push((rx, qp2p));
+        }
+
+        // Create a client and bootstrap to 4 of the peers we created previously.
+        let (mut qp2p0, rx0) = new_random_qp2p_for_unit_test(false, hcc_contacts);
+        qp2p0.bootstrap();
+
+        match unwrap!(rx0.recv()) {
+            Event::BootstrappedTo { node, .. } => {
+                assert!(bs_peer_addrs.contains(&node.peer_addr));
+            }
+            ev => panic!("Unexpected event: {:?}", ev),
+        }
+
+        thread::sleep(Duration::from_millis(if let Some(delay_ms) = delay_ms {
+            num_conns * delay_ms + delay_ms
+        } else {
+            100
+        }));
+
+        // We expect only 1 bootstrap connection to succeed.
+        let conns_count = unwrap!(qp2p0.connections(|c| {
+            c.iter()
+                .filter(|(_pa, conn)| conn.to_peer.is_established())
+                .count()
+        }));
+        assert_eq!(conns_count, 1);
+    }
+}
--- a/src/connection/bootstrap_group.rs
+++ b/src/connection/bootstrap_group.rs
@@ -14,10 +14,12 @@
 //! currently being made by the members of the group and thus an eventual destruction of all such
 //! members to not continue to use resources as we no longer require them.
 
+use crate::context::ctx_mut;
 use crate::event::Event;
 use crate::utils::ConnectTerminator;
 use std::cell::RefCell;
 use std::collections::HashMap;
+use std::mem;
 use std::net::SocketAddr;
 use std::rc::Rc;
 use std::sync::mpsc::Sender;
@@ -80,15 +82,29 @@ impl BootstrapGroupRef {
     /// is because the bootstrapping was successful (in which case no failure event will be
     /// auto-fired).
     pub fn terminate_group(&self, is_due_to_success: bool) {
-        let mut group = self.group.borrow_mut();
+        let mut terminators = {
+            let mut group = self.group.borrow_mut();
+            if is_due_to_success {
+                group.is_bootstrap_successful_yet = true;
+            }
 
-        if is_due_to_success {
-            group.is_bootstrap_successful_yet = true;
-        }
+            // We use a `mem::replace` here because `self.group` can be mutably borrowed
+            // twice during `BootstrapGroupRef::drop` (it's called when we remove a connection)
+            mem::replace(&mut group.terminators, Default::default())
+        };
 
-        for (_, mut terminator) in group.terminators.drain() {
-            let _ = terminator.try_send(());
-        }
+        let _ = terminators.remove(&self.peer_addr);
+
+        ctx_mut(|c| {
+            for (peer_addr, mut terminator) in terminators.drain() {
+                let _ = terminator.try_send(());
+                let _conn = c.connections.remove(&peer_addr);
+            }
+        });
+    }
+
+    pub fn is_bootstrap_successful_yet(&self) -> bool {
+        self.group.borrow().is_bootstrap_successful_yet
     }
 }
 
--- a/src/connection/from_peer.rs
+++ b/src/connection/from_peer.rs
@@ -63,7 +63,8 @@ impl fmt::Debug for FromPeer {
                 "FromPeer::Established with {} pending reads",
                 pending_reads.len()
             ),
-            ref blah => write!(f, "{:?}", blah),
+            FromPeer::NoConnection => write!(f, "FromPeer::NoConnection"),
+            FromPeer::NotNeeded => write!(f, "FromPeer::NotNeeded"),
         }
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -144,7 +144,13 @@ pub struct QuicP2p {
 }
 
 impl QuicP2p {
-    /// Bootstrap to a proxy
+    /// Bootstrap to the network.
+    ///
+    /// Bootstrap concept is different from "connect" in several ways: `bootstrap()` will try to
+    /// connect to all peers which are specified in the config (`hard_coded_contacts`) or were
+    /// previously cached. If one bootstrap connection succeeds, all other connections will be dropped.
+    ///
+    /// In case of success `Event::BootstrapedTo` will be fired. On error quic-p2p will fire `Event::BootstrapFailure`.
     pub fn bootstrap(&mut self) {
         self.el.post(|| {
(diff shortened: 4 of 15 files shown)


src/lib.rs
```diff
@@ -573,6 +578,70 @@ mod tests {
...
+
+        let (tx1, _rx1) = mpsc::channel();
+        let mut malicious_client = TestPeer::new(tx1);
+        unwrap!(malicious_client.activate(OurType::Client));
+        malicious_client.send(
+            qp2p0_info.clone().into(),
+            WireMsg::Handshake(Handshake::Node { cert_der: vec![] }),
+        );
+
+        std::thread::sleep(Duration::from_millis(10));
```
ustulation: Instead of delays can we not just do `recv()` below (followed by try_recv again ofc)

src/lib.rs
```diff
@@ -573,6 +578,70 @@ mod tests {
...
+            }) => {}
+            r => panic!("Unexpected result {:?}", r),
+        }
+
+        // No more messages expected
+        match rx0.try_recv() {
+            Err(TryRecvError::Empty) => {}
+            r => panic!("Unexpected result {:?}", r),
+        }
+    }
```
ustulation: At this stage also check that the peer-type hasn't changed in `qp2p0`

src/lib.rs
```diff
@@ -573,6 +578,70 @@ mod tests {
...
```
ustulation: At this stage also check that the peer-type hasn't changed in `qp2p0`

src/test_utils.rs
```diff
@@ -0,0 +1,203 @@
...
+            let _ = err_tx.send(Ok(()));
+        });
+
+        (err_rx.recv()?)?;
+
+        Ok(())
+    }
+
+    /// Send an arbitrary message.
+    pub fn send(&mut self, peer: Peer, msg: WireMsg) {
```
ustulation: If everything else is the same and only this is different can we now just _add_ this function for `QuicP2p` as an imp in tests ?

like in `test_utils`:
```
impl QuicP2p {
    fn send_wire_msg(...) { .. }
}
```

nbaksalyar: Ah that's a nice trick I totally forgot about. Yeah, it should work and we don't need an explicit `activate` fn taking `OurType` either because it could be always provided in the config. :+1:

src/bootstrap.rs
```diff
@@ -30,3 +30,60 @@ pub fn start() {
...
+            let (tx, rx) = mpsc::channel();
+            let mut qp2p = unwrap!(Builder::new(tx)
+                .with_config(Config {
+                    ip: Some(IpAddr::V4(Ipv4Addr::LOCALHOST)),
+                    port: Some(0),
+                    ..Default::default()
+                })
+                .build());
+
+            qp2p.set_connect_delay(i * 30);
```
ustulation: what are we testing with this delay ?

src/bootstrap.rs
```diff
@@ -30,3 +30,60 @@ pub fn start() {
...
+        qp2p0.bootstrap();
+
+        match unwrap!(rx0.recv()) {
+            Event::BootstrappedTo { node, .. } => {
+                assert_eq!(node.peer_addr, bs_peer_addrs[0]);
+            }
+            ev => panic!("Unexpected event: {:?}", ev),
+        }
+
+        // We expect only 1 bootstrap connection to succeed.
```
ustulation: Well you have put a delay for the other connections, so even if they matured and there was a bug in production to not eject them, the test would have ended by then. So i'm not getting what exactly is being tested.

src/connection/bootstrap_group.rs
```diff
@@ -80,15 +82,29 @@ impl BootstrapGroupRef {
...
+            mem::replace(&mut group.terminators, Default::default())
+        };
 
-        for (_, mut terminator) in group.terminators.drain() {
-            let _ = terminator.try_send(());
-        }
+        ctx_mut(|c| {
+            for (peer_addr, mut terminator) in terminators.drain() {
+                let _ = terminator.try_send(());
+                if peer_addr != self.peer_addr {
```
ustulation: Do we ever add ourselves to the map ? If we did that would be buggy, and even in that case we should promptly remove the connection anyway ?

src/connection/from_peer.rs
```diff
@@ -63,7 +63,8 @@ impl fmt::Debug for FromPeer {
                 "FromPeer::Established with {} pending reads",
                 pending_reads.len()
             ),
-            ref blah => write!(f, "{:?}", blah),
+            FromPeer::NoConnection => write!(f, "FromPeer::NoConnection"),
```
ustulation: lol, sorry

src/lib.rs
```diff
@@ -573,6 +585,114 @@ mod tests {
...
+
+        // No more messages expected
+        match rx0.try_recv() {
+            Err(TryRecvError::Empty) => {}
+            r => panic!("Unexpected result {:?}", r),
+        }
+
+        // Check that both have unchanged `ToPeer`/`FromPeer` types.
+        assert!(unwrap!(malicious_client.connections(move |c| {
+            if let FromPeer::Established { .. } = c[&qp2p0_info.peer_addr].from_peer {
```
ustulation: isn't there a function `is_established()` etc that you can use ?

src/lib.rs
```diff
@@ -573,6 +585,114 @@ mod tests {
...
+                true
+            } else {
+                false
+            }
+        })));
+
+        assert!(unwrap!(qp2p0.connections(move |c| {
+            if let ToPeer::Established { .. } = c[&malicious_client_info.peer_addr].to_peer {
```
ustulation: isn't there a function `is_established()` etc that you can use ?

src/lib.rs
```diff
@@ -573,6 +585,114 @@ mod tests {
...
+        assert!(unwrap!(malicious_client.connections(move |c| {
+            if let FromPeer::NotNeeded = c[&qp2p0_info.peer_addr].from_peer {
+                true
+            } else {
+                false
+            }
+        })));
+
+        assert!(unwrap!(qp2p0.connections(move |c| {
+            if let ToPeer::NotNeeded = c[&malicious_client_info.peer_addr].to_peer {
```
ustulation: use the function available in production code

src/lib.rs
```diff
@@ -573,6 +585,114 @@ mod tests {
...
+
+        // No more messages expected.
+        match rx0.try_recv() {
+            Err(TryRecvError::Empty) => {}
+            r => panic!("Unexpected result {:?}", r),
+        }
+
+        // Check that both have unchanged `ToPeer`/`FromPeer` types.
+        assert!(unwrap!(malicious_client.connections(move |c| {
+            if let FromPeer::NotNeeded = c[&qp2p0_info.peer_addr].from_peer {
```
ustulation: use the function available in production code

src/listener.rs
```diff
@@ -28,6 +29,12 @@ pub fn listen(incoming_connections: quinn::Incoming) {
     current_thread::spawn(leaf);
 }
 
+enum NewConnState {
+    Duplicate(QConn),
+    AlreadyBootstrapped,
+    Nothing(Option<BootstrapGroupRef>),
```
ustulation:
```suggestion
    Continue(Option<BootstrapGroupRef>),
```

src/listener.rs
```diff
@@ -76,17 +87,32 @@ fn handle_new_conn(
...
                     info!("ERROR in informing user about a new peer: {:?} - {}", e, e);
                 }
-            }
-            None
+
+                bootstrap_group
+            } else {
+                None
+            };
+            NewConnState::Nothing(bootstrap_group)
```
ustulation:
```suggestion
            NewConnState::Continue(bootstrap_group)
```

src/listener.rs
```diff
@@ -76,17 +87,32 @@ fn handle_new_conn(
...
+        NewConnState::Duplicate(_q_conn) => {
+            debug!("Not allowing duplicate connection from peer: {}", peer_addr);
+        }
+        NewConnState::AlreadyBootstrapped => crate::utils::handle_communication_err(
+            peer_addr,
+            &Error::ConnectionCancelled,
+            "Connection already bootstrapped",
+            None,
+        ),
+        NewConnState::Nothing(bootstrap_group) => {
```
ustulation:
```suggestion
        NewConnState::Continue(bootstrap_group) => {
```

src/listener.rs
```diff
@@ -28,6 +29,12 @@ pub fn listen(incoming_connections: quinn::Incoming) {
     current_thread::spawn(leaf);
 }
 
+enum NewConnState {
```
ustulation:
```suggestion
enum Action {
```

src/listener.rs
```diff
@@ -28,6 +29,12 @@ pub fn listen(incoming_connections: quinn::Incoming) {
...
+    Duplicate(QConn),
```
ustulation:
```suggestion
    HandleDuplicate(QConn),
```

src/listener.rs
```diff
@@ -28,6 +29,12 @@ pub fn listen(incoming_connections: quinn::Incoming) {
...
+    AlreadyBootstrapped,
```
ustulation:
```suggestion
    HandleAlreadyBootstrapped,
```

src/listener.rs
```diff
@@ -76,17 +87,32 @@ fn handle_new_conn(
...
-            None
+
+                bootstrap_group
+            } else {
+                None
+            };
+            NewConnState::Nothing(bootstrap_group)
         } else {
-            Some(q_conn)
+            NewConnState::Duplicate(q_conn)
```
ustulation:
```suggestion
            NewConnState::HandleDuplicate(q_conn)
```

src/listener.rs
```diff
@@ -53,19 +60,23 @@ fn handle_new_conn(
...
                     peer_cert_der: peer_cert_der.clone(),
                 };
+                let mut bootstrap_group = None;
 
                 // TODO come back to all the connected-to events and see if we are handling all
                 // cases
                 let event = if let Some(bootstrap_group_ref) = conn.bootstrap_group_ref.take() {
-                    bootstrap_group_ref.terminate_group(true);
+                    if bootstrap_group_ref.is_bootstrap_successful_yet() {
+                        return NewConnState::AlreadyBootstrapped;
```
ustulation:
```suggestion
                        return NewConnState::HandleAlreadyBootstrapped;
```

src/listener.rs
```diff
@@ -76,17 +87,32 @@ fn handle_new_conn(
...
-            Some(q_conn)
+            NewConnState::Duplicate(q_conn)
         }
     });
 
-    if let Some(_q_conn) = is_duplicate {
-        debug!("Not allowing duplicate connection from peer: {}", peer_addr);
-        return;
+    match state {
+        NewConnState::Duplicate(_q_conn) => {
```
ustulation:
```suggestion
        NewConnState::HandleDuplicate(_q_conn) => {
```

src/listener.rs
```diff
@@ -76,17 +87,32 @@ fn handle_new_conn(
...
+            debug!("Not allowing duplicate connection from peer: {}", peer_addr);
+        }
+        NewConnState::AlreadyBootstrapped => crate::utils::handle_communication_err(
```
ustulation:
```suggestion
        NewConnState::HandleAlreadyBootstrapped => crate::utils::handle_communication_err(
```

src/connection/bootstrap_group.rs
nbaksalyar: >If we did that would be buggy, and even in that case we should promptly remove the connection anyway ?
Hmm, not sure what you mean exactly. This list of connections is taken directly from `ctx_mut`, so naturally it will contain _all_ connections that've been made, including a succeeding bootstrapped connection (which is `self.peer_addr`). Why would we require to terminate it?

src/bootstrap.rs
```diff
@@ -35,24 +35,26 @@ pub fn start() {
...
     // Try to bootstrap to 4 different peer at once. Use an artificial delay for 3 of the peers.
     // Make sure in the end we have only one bootstrapped connection (all other connections are
     // supposed to be dropped).
     #[test]
     fn bootstrap_to_only_one_node() {
-        env_logger::init();
+        const NUM_CONNS: u64 = 4;
+        const DELAY_MS: u64 = 10;
 
         let mut bs_nodes = Vec::new();
```
ustulation: Normally since this is a test case, i woulnd't have put the review, but i don't want to merge PR with some of the commits saying "review fixes" so this needs rebased (or squashed) anyway so was thinking i'll just mention this review as well if you want to incorporate that:
Since now you have constants for `NUM_CONNS` etc. which is good, why not also do `with_capacity()` for the containers which allow it ? Not critical for test cases but if other changes are there the we can just do these too.

src/bootstrap.rs
```diff
@@ -30,3 +30,64 @@ pub fn start() {
...
+        qp2p0.bootstrap();
+
+        match unwrap!(rx0.recv()) {
+            Event::BootstrappedTo { node, .. } => {
+                assert_eq!(node.peer_addr, bs_peer_addrs[0]);
+            }
+            ev => panic!("Unexpected event: {:?}", ev),
+        }
+
+        thread::sleep(Duration::from_millis(NUM_CONNS * DELAY_MS));
```
ustulation:
```suggestion
        thread::sleep(Duration::from_millis(NUM_CONNS * DELAY_MS + DELAY_MS));
```

To make sure we wait a little bit more than the exact delay for all the connection attempts

src/bootstrap.rs
```diff
@@ -30,3 +30,64 @@ pub fn start() {
...
+            let (tx, rx) = mpsc::channel();
+            let mut qp2p = unwrap!(Builder::new(tx)
+                .with_config(Config {
+                    ip: Some(IpAddr::V4(Ipv4Addr::LOCALHOST)),
+                    port: Some(0),
+                    ..Default::default()
+                })
+                .build());
+
+            qp2p.set_connect_delay(i * DELAY_MS);
```
ustulation: In fact just to be robust, i would say have a few connect without delay too (so simultaneous) and make sure that we still get one connection. You can make this a private function taking delay as optional parameter. If None then set it to zero else use the supplied one - that way you will have two test cases - one this exact one simulated and the other without delay.

If the Delay is 0 then ofc sleep for some non-zero interval in the code below

PR #44, opened 2019-04-26
return unsent messages

ustulation:
User messages unable to be delivered will be returned back to the user.
Currently this will happen only in cases where there was a prior connection already and they attempted to write something to it.
All messages falied to be sent would be returned.

A future work to this PR should also see even those messages to be returned when connection attempt failed or the user invoked an illegal API (e.g., trying to send to an unconnected client where no connection attempt is made on send - this attempt is made on.y for nodes because clients are deemed not reverse connectible). Currently these two case would swallow the message. However for the rest of (majority of) cases, the unsent user messages would be returned).

Also note that without App level ACKs there is no guaranteed way to know if these failed. In most cases when we get an error while sending it would actually mean that the message was not entirely sent, so likely discarded at the peer end. However due to network anomalies it might happen that it was sent and then some disruption happened which prevented us from knowing the message was actually delivered. Again, these would be minority of the cases and if such robustness is required we should use App level ACKs which quic-p2p doesn't currently have.

### DIFF
--- a/src/communicate.rs
+++ b/src/communicate.rs
@@ -109,19 +109,37 @@ pub fn write_to_peer(peer_addr: SocketAddr, msg: WireMsg) {
 
 /// Write to the peer, given the QUIC connection to it
 pub fn write_to_peer_connection(peer_addr: SocketAddr, conn: &QConn, wire_msg: WireMsg) {
+    let user_msg = if let WireMsg::UserMsg(ref m) = wire_msg {
+        Some(m.clone())
+    } else {
+        None
+    };
+    let user_msg0 = user_msg.clone();
+    let user_msg1 = user_msg.clone();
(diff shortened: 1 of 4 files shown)


dirvine:
Nice on @ustulation 
>  However due to network anomalies it might happen that it was sent and then some disruption happened which prevented us from knowing the message was actually delivered. Again, these would be minority of the cases and if such robustness is required we should use App level ACKs which quic-p2p doesn't currently have.
We won't go down that path, we will treat this as best endeavour and not all edge cases catered for, that would be months of work and may still leave the door open for weird cases, so msg got delivered but did it get processed etc. etc. if a quic-p2p consumer needs that then they need to provide a mechanism higher up that us here :+1:

PR #56, opened 2019-06-20
fix: handle key and certificate parse errors

d1vyank:
Addresses a TODO to handle an error from parsing key material

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -106,12 +106,11 @@ pub struct SerialisableCertificate {
 }
 
 impl SerialisableCertificate {
-    // TODO do proper error handling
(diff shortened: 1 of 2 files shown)


ustulation:
This was marked TODO because it could be read from the disk which could contain invalid stuff which will cause an error. Although I'm not clear if panicking is unwarranted in these situations as it's really user's fault to have put something invalid in the config and since that's the start-up scenario we should bail out as a hard error, I'm still inclined to accept this and pass the buck to the user library to deal with it. Maybe they want to fallback to something else on error etc.

So thanks for the PR; going to merge it now !

PR #57, opened 2019-06-21
Enable standard lints and fix resulting issues

### LINKED ISSUE #55: Enable standard lints
We should enable the standard set of lint settings and resolve the resulting errors.

d1vyank:
Adds standard lint rules used by the routing and parsec crates.

Fixes #55

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,9 +16,8 @@ before_script:
     fi
   - rustup component add rustfmt clippy
 script:
-  - cargo fmt -- --check
   - if [ "${TRAVIS_OS_NAME}" = linux ]; then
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -18,7 +18,7 @@ before_script:
 script:
   - cargo fmt -- --check
```
nbaksalyar: I think this line can be removed now because it's included in `scripts/clippy`? And in that case it's probably better to rename this script because it also does the formatting.

Alternatively, it can be removed from `scripts/clippy` :)

d1vyank: Done, took the first suggestion!

PR #59, opened 2019-06-28
fix: return error on connection failure

d1vyank:
Addresses a TODO to return an error on outgoing connection failure.


ustulation:
Thanks ! Looks good.

PR #62, opened 2019-07-15
changes due to routing requirements

ustulation:
Take a token from the user and return it during, both, send successes and failures, along with the message and the peer that the send attempt was made to

### DIFF
--- a/src/communicate.rs
+++ b/src/communicate.rs
@@ -12,7 +12,7 @@ use crate::connection::{Connection, FromPeer, QConn, ToPeer};
 use crate::context::{ctx, ctx_mut};
 use crate::error::Error;
 use crate::event::Event;
-use crate::utils;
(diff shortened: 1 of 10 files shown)


src/communicate.rs
```diff
@@ -143,7 +149,7 @@ pub fn write_to_peer_connection(peer_addr: SocketAddr, conn: &QConn, wire_msg: W
                 )
             })
         })
-        .map(|_| ());
+        .map(move |_| utils::handle_send_success(peer_addr, user_msg2));
```
jeanphilippeD: Will that do the right thing?
i.e will we not fail to send the success message if `tokio::io::write_all` return None? or tokio shutdown.

jeanphilippeD: My bad I guess, in these case it will call handle_communication_err

PR #69, opened 2019-08-15
fix/wire_msg: fix incorrect serialisation logic in WireMsg

nbaksalyar:
While deserialising messages, we wrongly assumed that all messages >1 KB in size are not serialised and are plain user messages. However, `bincode::serialise(wire_msg)` also could have produced messages over 1 KB, which resulted in deserialisation failures in the user code.

To fix this, we added a one-byte flag that will clearly identify pre-serialised user messages and messages that are sent by quic-p2p itself. With that, the serialisation logic becomes even more efficient now because we don't need to serialise _all_ user messages which are likely to be serialised already.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -85,16 +85,20 @@ quick_error! {
          OperationNotAllowed {
              display("This operation is not allowed for us")
          }
-        /// Connection Cancelled
-        ConnectionCancelled {
-            display("Connection was actively cancelled")
-            from()
-        }
-        /// Failed receiving from an `mpsc::channel`.
-        ChannelRecv(e: mpsc::RecvError) {
-            display("Channel receive error: {}", e)
-            cause(e)
-            from()
-        }
+         /// Connection Cancelled
+         ConnectionCancelled {
+             display("Connection was actively cancelled")
+             from()
+         }
+         /// Failed receiving from an `mpsc::channel`.
+         ChannelRecv(e: mpsc::RecvError) {
+             display("Channel receive error: {}", e)
+             cause(e)
+             from()
+         }
+         /// Could not deserialise a wire message.
+         InvalidWireMsgFlag {
+             display("Could not deserialise a wire message: unexpected message type flag")
+         }
      }
 }
--- a/src/wire_msg.rs
+++ b/src/wire_msg.rs
@@ -7,11 +7,9 @@
 // specific language governing permissions and limitations relating to use of the SAFE Network
 // Software.
 
-use crate::{utils, R};
+use crate::{error::Error, utils, R};
 use std::fmt;
 use std::net::SocketAddr;
-
-const MAX_MESSAGE_SIZE_FOR_SERIALISATION: usize = 1024; // 1 KiB
 
 /// Final type serialised and sent on the wire by QuicP2p
 #[derive(Serialize, Deserialize, Debug)]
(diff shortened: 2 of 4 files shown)


src/wire_msg.rs
```diff
@@ -22,25 +20,27 @@ pub enum WireMsg {
...
-            }
-        }
+const USER_MESSAGE: u8 = 0;
+const WIRE_MESSAGE: u8 = 1;
 
-        From::from(unwrap!(bincode::serialize(&self)))
+impl Into<(u8, bytes::Bytes)> for WireMsg {
+    fn into(self) -> (u8, bytes::Bytes) {
+        match self {
+            WireMsg::UserMsg(ref m) => (USER_MESSAGE, m.clone()),
```
ustulation: `From` and `Into` are asymmetric. Maybe better to return `(m.clone, USER_MSG_FLAG)` since the flag is the last byte. For the other branch it will ofc be `(serialize(...), !USER_MSG_FLAG)`.

src/wire_msg.rs
```diff
@@ -22,25 +20,27 @@ pub enum WireMsg {
...
 
-impl Into<bytes::Bytes> for WireMsg {
-    fn into(self) -> bytes::Bytes {
-        if let WireMsg::UserMsg(ref m) = self {
-            if m.len() > MAX_MESSAGE_SIZE_FOR_SERIALISATION {
-                return m.clone();
-            }
-        }
+const USER_MESSAGE: u8 = 0;
+const WIRE_MESSAGE: u8 = 1;
```
ustulation: Even a user msg is a wire msg, so this naming convention is a bit strange.
```suggestion
const USER_MSG_FLAG: u8 = 0;
```
Then for non-user-msg, just do `!USER_MSG_FLAG`. So no need of two variants.

src/wire_msg.rs
```diff
@@ -22,25 +20,27 @@ pub enum WireMsg {
...
 impl WireMsg {
-    pub fn from_raw(raw: Vec<u8>) -> R<Self> {
-        if raw.len() > MAX_MESSAGE_SIZE_FOR_SERIALISATION {
-            return Ok(WireMsg::UserMsg(From::from(raw)));
-        }
+    pub fn from_raw(mut raw: Vec<u8>) -> R<Self> {
+        let is_user_message = raw.pop();
 
-        Ok(bincode::deserialize(&raw)?)
+        match is_user_message {
```
ustulation:
```suggestion
        match msg_flag {
            Some(USER_MSG_FLAG) => Ok(WireMsg::UserMsg(From::from(raw))),
            Some(!USER_MSG_FLAG) => Ok(bincode::deserialize(&raw)?),
            _x => Err(Error::InvalidWireMsgFlag),
        }
```

src/wire_msg.rs
```diff
@@ -22,25 +20,27 @@ pub enum WireMsg {
...
     }
 }
 
 impl WireMsg {
-    pub fn from_raw(raw: Vec<u8>) -> R<Self> {
-        if raw.len() > MAX_MESSAGE_SIZE_FOR_SERIALISATION {
-            return Ok(WireMsg::UserMsg(From::from(raw)));
-        }
+    pub fn from_raw(mut raw: Vec<u8>) -> R<Self> {
+        let is_user_message = raw.pop();
```
ustulation:
```suggestion
        let msg_flag = raw.pop();
```

src/error.rs
```diff
@@ -85,16 +85,20 @@ quick_error! {
...
+             from()
+         }
+         /// Failed receiving from an `mpsc::channel`.
+         ChannelRecv(e: mpsc::RecvError) {
+             display("Channel receive error: {}", e)
+             cause(e)
+             from()
+         }
+         /// Could not deserialise a wire message.
+         InvalidWireMsg {
```
ustulation:
```suggestion
         InvalidWireMsgFlag {
```

PR #76, opened 2019-10-14
Allow specifying a custom directory for the boostrap cache

### LINKED ISSUE #73: qp2p::dir::Dirs not public -> Config::read_or_construct_default(Option<Dirs>) is not usable
Hey,
I am now trying to setup multiple nodes with slightly different, non-default setups where I want to save the cert,...
When trying to use Config::read_or_construct_default I realized that I couldn't specify the Dirs as it is inside a private/pub(crate) module and isn't made public in lib.rs too.
Is this a bug or is there a different intended way to use this function?

Thanks in advance,
Chris

lionel1704:
Also fixes #73

### DIFF
--- a/src/dirs.rs
+++ b/src/dirs.rs
@@ -33,9 +33,13 @@ impl OverRide {
     }
 }
 
+/// Location at which data such as config, keys, certificates and other backup data
+/// are stored.
 pub enum Dirs {
+    /// User defined custom directory for all data.
     #[allow(unused)]
     Overide(OverRide),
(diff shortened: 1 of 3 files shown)


src/dirs.rs
```diff
@@ -43,6 +46,7 @@ pub enum Dirs {
         ),
         windows
     ))]
+    /// Set of directories to specifically store the cache, config app data etc.
```
ustulation: Doc comments should precede the attributes (like for the other variant you did). That's the convention we have.

PR #80, opened 2019-12-07
Remove confusing proxy name

dirvine:
As we do not have proxies this name could only confuse.

PR #82, opened 2019-12-22
Use quinn 0.5.0 and async/await code

nbaksalyar:
Also upgrade `tokio` to 0.5.2 and `futures` to 0.3.
Supersedes and includes changes from #80 and #79.


src/communicate.rs
```diff
@@ -145,96 +145,110 @@ pub fn write_to_peer_connection(
...
-            })
-        });
+            return;
+        }
+
+        utils::handle_send_success(peer_addr, user_msg);
+    };
 
-    current_thread::spawn(leaf);
+    let _ = tokio::spawn(leaf);
```
dirvine: Is this _ pattern required for tokio?

src/communicate.rs
```diff
@@ -145,96 +145,110 @@ pub fn write_to_peer_connection(
...
+    peer_addr: SocketAddr,
+    mut uni_streams: quinn::IncomingUniStreams,
+    mut bi_streams: quinn::IncomingBiStreams,
+) {
+    let _ = tokio::spawn(async move {
+        if let Some(res) = bi_streams.next().await {
+            let err = match res {
+                Err(e) => {
+                    debug!(
+			"Error in Incoming-bi-stream while reading from peer {}: {:?} - {}.\nNote: It
```
dirvine: Although bi-directional are allowed for clients? Perhaps reword to make that clear?

dirvine: i.e. p2p nodes cannot be connected bi-directionally. Client connections are bi-directional, but p2p nodes cannot be.

src/config.rs
```diff
@@ -118,19 +119,26 @@ impl SerialisableCertificate {
...
+            quinn::Certificate::from_der(&self.cert_der)
+                .map_err(|e| Error::CertificateParseError(format!("{:?}, {}", e, e)))?,
         ))
     }
 }
 
 impl Default for SerialisableCertificate {
     fn default() -> Self {
-        let cert = rcgen::generate_simple_self_signed(vec!["MaidSAFE.net".to_string()]);
+        let cert = unwrap!(rcgen::generate_simple_self_signed(vec![
```
dirvine: We need to handle this error.

src/config.rs
```diff
@@ -118,19 +119,26 @@ impl SerialisableCertificate {
...
+            "MaidSAFE.net".to_string()
+        ]));
 
         Self {
-            cert_der: cert.serialize_der(),
+            cert_der: unwrap!(cert.serialize_der()),
```
dirvine: Same as above.

src/connection/mod.rs
```diff
@@ -123,12 +117,10 @@ fn spawn_incomplete_conn_killer(peer_addr: SocketAddr) {
                     let _ = conn.remove();
                 }
             });
-
-            Ok(())
         });
 
     // TODO find a way to cancel this timer if we know the connection is done. Otherwise it
```
dirvine: Can we make sure this is a GH issue and give it some priority

src/error.rs
```diff
@@ -100,5 +100,21 @@ quick_error! {
          InvalidWireMsgFlag {
              display("Could not deserialise a wire message: unexpected message type flag")
          }
+         /// Write error
```
dirvine: I will do a follow-up PR to remove quick_error and we will use err-derive as quinn does. That should give us composable errors based on std::Error

src/event_loop.rs
```diff
@@ -64,21 +62,24 @@ pub struct EventLoop {
...
+                        if let Some(mut f) = ev_loop_msg.0 {
+                            f();
+                        } else {
+                            break;
+                        }
                     }
-                });
+                };
+
+                let mut runtime = unwrap!(tokio::runtime::Runtime::new());
```
dirvine: Here if we cannot start runtime we may need to panic, but with a clear message. We can add description to unwrap here AFAIK `unwrap!(tokio::runtime::Runtime::new(), "Cannot start quick-p2p runtime, aborting")

src/peer_config.rs
```diff
@@ -25,15 +26,21 @@ pub const DEFAULT_IDLE_TIMEOUT_MSEC: u64 = 30_000; // 30secs
...
+        .map_err(|e| Error::CertificateParseError(format!("{:?}, {}", e, e)))?;
 
     let mut peer_cfg_builder = {
         let mut client_cfg = quinn::ClientConfig::default();
         client_cfg.transport = Arc::new(new_transport_cfg(None, None));
 
         quinn::ClientConfigBuilder::new(client_cfg)
     };
-    let _ = peer_cfg_builder.add_certificate_authority(peer_cert)?;
+    let _ = peer_cfg_builder
```
dirvine: We should handle this result.

dirvine (commented):
Looks great, just some concern with handling result, esp where we spawn async workers.

src/communicate.rs
octol: Seems like there are some tab whitespace here?

src/connect.rs
```diff
@@ -79,31 +83,26 @@ pub fn connect_to(
...
+                select! {
+                // Terminator leaf
+                            _ = rx.recv().fuse() => {
+                                handle_connect_err(peer_addr, &Error::ConnectionCancelled);
+                            },
+                            // New connection
+                            new_peer_conn_res = connecting.fuse() => {
+                                handle_new_connection_res(peer_addr, new_peer_conn_res);
+                            },
+                    }
```
octol: Nitpick: indentation?

src/communicate.rs
nbaksalyar: Yes, [`tokio::spawn`](https://docs.rs/tokio/0.2.8/tokio/fn.spawn.html) returns a `JoinHandle` now which is similar to a handle returned by `thread::spawn`. We don't have a use for it, hence `let _ = ...`

src/communicate.rs
nbaksalyar: >Although bi-directional are allowed for clients?

Not quite, it's a bit confusing thing, but we don't use bi-directional streams at all. For clients it's more like 2 uni-directional streams instead of 1 bi-directional.

src/config.rs
nbaksalyar: Might need a slight code rewrite here because this one is in the `Default` impl which doesn't return a `Result`, so I'd prefer to tackle it in a separate PR along with quick_error removal if you don't mind?

src/peer_config.rs
nbaksalyar: This one should be handled: it goes like `let _ = peer_cfg_builder.map_err(..)?`, so the discarded result here is only `Ok(ConfigBuilder)` which we have no use for.

octol (approved):

PR #87, opened 2020-01-08
Mock

madadam:
This PR adds mock version of quic-p2p as a workspace subcrate. The code itself is extracted from routing.


mock/src/lib.rs
```diff
@@ -0,0 +1,349 @@
...
+            "client" => Ok(Self::Client),
+            "node" => Ok(Self::Node),
+            x => Err(format!("Unknown client type: {}", x)),
+        }
+    }
+}
+
+/// Events from `QuicP2p` to the user.
+#[derive(Debug)]
+pub enum Event {
```
jeanphilippeD: Since it is in the same repo, is it worth using the real event?

jeanphilippeD: No, I guess since it does not have a dependency on quic_p2p, make sense.

jeanphilippeD (commented):
Looks good to me assuming the routing part make sense too.

mock/src/lib.rs
madadam: Well, currently mock doesn't depend on the real thing, but maybe it should? Then we would be able to reuse some of the definitions. That could also make it a bit easier to ensure both APIs remain identical. What do you think?

jeanphilippeD:
Seems like you need to rust-fmt it and then I'll approve.

jeanphilippeD (dismissed):
To merge once fmt is fixed

mock/src/lib.rs
nbaksalyar: >That could also make it a bit easier to ensure both APIs remain identical

This would make total sense, especially if we start to change the code to use async/await features. However, I think we can create an issue for now and tackle this later.

nbaksalyar (dismissed):
Looks good to be merged after the rustfmt is fixed :+1:

jeanphilippeD:
Branch looks out of date with master, seem it needs rebase?

nbaksalyar (approved):

jeanphilippeD (approved):

PR #89, opened 2020-01-31
Update for rust 1.41 (mem::replace -> mem::take))

dirvine:
Replace quick_error with err_derive to be in line with quinn routing and also
derviced from std::Error
Removed format trick from CertificateError and created an issue in quinn to make tls
pub and allow access to those errors for composability.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -50,6 +50,8 @@ pub enum QuicP2pError {
     ConnectionCancelled,
     #[error(display = "Channel receive error ")]
     ChannelRecv(#[source] mpsc::RecvError),
+    #[error(display = "Could not add certificate to PKI")]
+    WebPki,
     #[error(display = "Invalid wire message.")]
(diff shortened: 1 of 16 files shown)


src/error.rs
```diff
@@ -7,115 +7,57 @@
...
-             from()
-         }
-         /// Add certificate error
-         AddCertificateError(e: String) {
-             display("Could not add certificate: {}", e)
-         }
-     }
+#[derive(Debug, Error)]
+#[allow(missing_docs)]
+pub enum QuicP2pError {
```
madadam: Why the rename?

madadam (commented):
Looks good. Just not sure what is the reason for the `Error` -> `QuicP2pError` rename.

src/error.rs
dirvine: Error was clashing with the `err_derive::Error`

PR #90, opened 2020-02-04
Add full peer info to all events

madadam:
This PR changes `Event` so it always includes the full peer info instead of just their socket address. This makes life easier for the consumers of this library because they no longer need to maintain separate mapping between socket address and peer info.

It also changes the representation of the serialized certificate from `Vec<u8>` to `Bytes` which makes is cheap to clone.

### DIFF
--- a/mock/src/tests.rs
+++ b/mock/src/tests.rs
@@ -549,7 +549,7 @@ impl Agent {
     ) {
         let (actual_addr, actual_msg, actual_id) = assert_match!(
             self.rx.try_recv(),
-            Ok(Event::SentUserMessage { peer_addr, msg, token }) => (peer_addr, msg, token)
+            Ok(Event::SentUserMessage { peer, msg, token }) => (peer.peer_addr(), msg, token)
         );
 
         assert_eq!(actual_addr, *dst_addr);
--- a/mock/src/tests.rs
+++ b/mock/src/tests.rs
@@ -566,7 +566,7 @@ impl Agent {
     ) {
         let (actual_addr, actual_msg, actual_id) = assert_match!(
             self.rx.try_recv(),
-            Ok(Event::UnsentUserMessage { peer_addr, msg, token }) => (peer_addr, msg, token)
+            Ok(Event::UnsentUserMessage { peer, msg, token }) => (peer.peer_addr(), msg, token)
         );
 
         assert_eq!(actual_addr, *dst_addr);
--- a/mock/src/tests.rs
+++ b/mock/src/tests.rs
@@ -525,7 +525,7 @@ impl Agent {
     fn expect_connection_failure(&self, addr: &SocketAddr) {
         let actual_addr = assert_match!(
             self.rx.try_recv(),
-            Ok(Event::ConnectionFailure { peer_addr, .. }) => peer_addr
+            Ok(Event::ConnectionFailure { peer, .. }) => peer.peer_addr()
         );
         assert_eq!(actual_addr, *addr);
     }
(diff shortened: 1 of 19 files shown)


src/communicate.rs
```diff
@@ -88,12 +93,15 @@ pub fn try_write_to_peer(peer: Peer, msg: WireMsg, token: Token) -> R<()> {
 
 /// This will fail if we don't have a connection to the peer or if the peer is in an invalid state
 /// to be sent a message to.
-pub fn write_to_peer(peer_addr: SocketAddr, msg: WireMsg, token: Token) -> R<()> {
+pub fn write_to_peer(peer: Peer, msg: WireMsg, token: Token) -> R<()> {
```
dirvine: Should this not be private and called from `write_to_peer_connection` So we always need to pass socketaddr, but if we are already connected then this method runs, otherwise we create a connection.

mock/src/tests.rs
```diff
@@ -259,6 +259,8 @@ fn send_to_disconnecting_node() {
     let mut a = Agent::node();
     let mut b = Agent::node();
 
+    println!("A={}, B={}", a.addr(), b.addr());
```
jeanphilippeD: is that suppose to be a println?

jeanphilippeD (dismissed):
Looks good and self contain. assuming Routing PR works out.

mock/src/tests.rs
madadam: Oops, must have left it there by accident. Will remove.

src/communicate.rs
madadam: Could be, but this doesn't seem to be related to the purpose of this PR, so I'd prefer to not do it here.

src/event.rs
```diff
@@ -58,49 +57,36 @@ pub enum Event {
...
-            Event::ConnectionFailure {
-                ref peer_addr,
-                ref err,
-            } => write!(
+            Event::ConnectionFailure { peer, err } => write!(
                 f,
-                "Event::ConnectionFailure {{ peer_addr: {}, err: {} }}",
-                peer_addr, err
+                "Event::ConnectionFailure {{ peer: {}, err: {} }}",
+                peer.peer_addr(),
```
nbaksalyar: Is there a reason to display only the address specifically? Given that `Peer` implements `Display` maybe it will be more informative to display it entirely, with the metainformation like `Peer::Node` or `Peer::Client`.

src/event.rs
```diff
@@ -58,49 +57,36 @@ pub enum Event {
...
-                ref peer_addr,
-                ref msg,
-                ref token,
-            } => write!(
+            Event::SentUserMessage { peer, msg, token } => write!(
                 f,
-                "Event::SentUserMessage {{ peer_addr: {}, msg: {}, token: {} }}",
-                peer_addr,
+                "Event::SentUserMessage {{ peer: {}, msg: {}, token: {} }}",
+                peer.peer_addr(),
```
nbaksalyar: Same here and below in all occurrences of `peer.peer_addr()`

src/lib.rs
```diff
@@ -197,15 +198,16 @@ impl QuicP2p {
 
     /// Connect to the given peer. This will error out if the peer is already in the process of
```
nbaksalyar: Might be worth also changing this comment to `Connect to the given node ...`

nbaksalyar (commented):
Looks good, great work! 👍
I have just some minor comments which you can address in a separate PR if preferred.

nbaksalyar (approved):

src/event.rs
madadam: I guess just to preserve the previous behavior. But probably makes sense to do what you say.

PR #91, opened 2020-02-05
Remove peer certificates 

madadam:
Currently in order to establish secure connection to a node, we need to obtain their certificate via a side-channel and then pass it to quic-p2p in the `NodeInfo` struct. Quic-p2p then effectively creates a separate "certificate authority" for each node using this certificate. This is insecure and inconvenient. This PR removes this mechanism by effectively disabling any certificate validation. This makes the API simpler but requires the consumers of quic-p2p to handle any authentication themselves. 

At some point in the future we will reintroduce certificate validation back, but likely using a proper certification authority.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,11 +22,12 @@ serde = { version = "~1.0.91", features = ["derive"] }
 serde_json = "~1.0.39"
 structopt = "~0.2.15"
 rcgen = "~0.7.0"
-rustls = "~0.16.0"
+rustls = { version = "~0.16.0", features = ["dangerous_configuration"] }
 log = "~0.4.6"
 base64 = "~0.10.1"
 err-derive = "^0.2.2"
 derive_more = "^0.99.2"
+webpki = "~0.21.2"
 
 [dev-dependencies]
 clap = "~2.32.0"
--- a/examples/chat.rs
+++ b/examples/chat.rs
@@ -12,13 +12,15 @@
 use bytes::Bytes;
 use crossbeam_channel as mpmc;
 use quic_p2p::{Builder, Config, Event, Peer, QuicP2p};
-use rand::{self, RngCore};
+use rand::{distributions::Standard, Rng};
 use rustyline::config::Configurer;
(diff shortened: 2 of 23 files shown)


examples/chat.rs
```diff
@@ -205,17 +205,14 @@ fn handle_qp2p_events(
...
             println!("Error getting ourinfo: {}", e);
             return;
         }
     };
 
-    println!(
-        "Our info:\n\n{}\n",
-        serde_json::to_string(&ourinfo).unwrap()
-    );
+    println!("Our info: {}", ourinfo);
```
nbaksalyar: This should to be
```suggestion
    println!("Our info: {}", ourinfo.peer_addr());
```
and the decoding part should be changed accordingly - currently, it expects a JSON-encoded `Peer` struct and the example doesn't work.

Optionally, this can also be encoded with `serde_json::to_string(...)` to simplify decoding.

nbaksalyar (changes requested):

examples/chat.rs
madadam: Thanks, fixed now.

jeanphilippeD:
Failure on windows @madadam ?

Cargo.toml
```diff
@@ -22,11 +22,12 @@ serde = { version = "~1.0.91", features = ["derive"] }
 serde_json = "~1.0.39"
 structopt = "~0.2.15"
 rcgen = "~0.7.0"
-rustls = "~0.16.0"
+rustls = { version = "~0.16.0", features = ["dangerous_configuration"] }
```
jeanphilippeD: That seem unfortunate in a production configuration.

madadam: I don't think the "dangerous_configuration" means "don't use this in production". It rather means "you should know what you are doing".

madadam:
> Failure on windows @madadam ?

Not just windows. Fixed now.

nbaksalyar (approved):
LGTM :+1: Great work, thanks!

PR #93, opened 2020-02-11
(WIP, Don't merge yet) Implement delayed packets in mock-quic-p2p

madadam:
The packets that indicate some kind of failure (`BootstrapFailure`, `ConnectFailure` and `MessageFailure`) are now delayed and delivered only after some time. This is not base on the actual wall-clock time, but instead is emulated by a call to `Network::advance_time` which the consumer of the library can call at appropriate times according to their needs.

This feature is needed by routing to be able to test node drops after the change to make drop detection more robust.

### DIFF
--- a/mock/src/network.rs
+++ b/mock/src/network.rs
@@ -49,6 +49,11 @@ impl Network {
         self.inner.borrow_mut().gen_addr(None, None)
     }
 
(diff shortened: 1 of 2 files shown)


jeanphilippeD (dismissed):

madadam:
Not needed. Closing.

PR #95, opened 2020-02-11
Take 2 channel one for client event one for node

jeanphilippeD:
Address the quic-p2p work for https://github.com/maidsafe/routing/issues/2039

The aim is to allow Vault to pass in a channel for its client events.
`EventReceivers` was made explicitly not part of the API as the 2 channels will be held by different objects Vault and Routing.

### DIFF
--- a/examples/common/mod.rs
+++ b/examples/common/mod.rs
@@ -27,12 +27,48 @@ pub struct EventReceivers {
 impl EventReceivers {
     #[allow(unused)]
     pub fn recv(&self) -> Result<Event, mpmc::RecvError> {
-        self.node_rx.recv()
+        let mut sel = mpmc::Select::new();
+        let client_idx = sel.recv(&self.client_rx);
+        let node_idx = sel.recv(&self.node_rx);
+        let selected_operation = sel.ready();
+
+        if selected_operation == client_idx {
+            self.client_rx.recv()
+        } else if selected_operation == node_idx {
+            self.node_rx.recv()
+        } else {
+            panic!("invalid operation");
+        }
     }
 
     #[allow(unused)]
-    pub fn iter(&self) -> mpmc::Iter<Event> {
-        self.node_rx.iter()
+    pub fn iter(&self) -> IterEvent {
+        IterEvent { event_rx: &self }
+    }
+}
+
+pub struct IterEvent<'a> {
+    event_rx: &'a EventReceivers,
+}
+
+impl<'a> Iterator for IterEvent<'a> {
+    type Item = Event;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let mut sel = mpmc::Select::new();
+        let client_idx = sel.recv(&self.event_rx.client_rx);
+        let node_idx = sel.recv(&self.event_rx.node_rx);
+        let selected_operation = sel.ready();
+
+        let event = if selected_operation == client_idx {
+            self.event_rx.client_rx.recv()
+        } else if selected_operation == node_idx {
+            self.event_rx.node_rx.recv()
+        } else {
+            return None;
+        };
+
+        event.ok()
     }
 }
 
--- a/examples/common/mod.rs
+++ b/examples/common/mod.rs
@@ -7,7 +7,8 @@
 // specific language governing permissions and limitations relating to use of the SAFE Network
 // Software.
 
-use quic_p2p::Peer;
+use crossbeam_channel as mpmc;
+use quic_p2p::{Event, EventSenders, Peer};
 use serde::{Deserialize, Serialize};
 
 /// Remote procedure call for our examples to communicate.
@@ -15,4 +16,32 @@ use serde::{Deserialize, Serialize};
 pub enum Rpc {
     /// Starts the connectivity and data exchange test between us and given QuicP2p peers.
     StartTest(Vec<Peer>),
+}
+
+/// Receivers side for events
+pub struct EventReceivers {
+    pub node_rx: mpmc::Receiver<Event>,
+    pub client_rx: mpmc::Receiver<Event>,
+}
+
+impl EventReceivers {
+    #[allow(unused)]
+    pub fn recv(&self) -> Result<Event, mpmc::RecvError> {
+        self.node_rx.recv()
+    }
+
+    #[allow(unused)]
+    pub fn iter(&self) -> mpmc::Iter<Event> {
+        self.node_rx.iter()
+    }
+}
+
+/// Create channels for events
+pub fn new_unbounded_channels() -> (EventSenders, EventReceivers) {
+    let (client_tx, client_rx) = mpmc::unbounded();
+    let (node_tx, node_rx) = mpmc::unbounded();
+    (
+        EventSenders { node_tx, client_tx },
+        EventReceivers { node_rx, client_rx },
+    )
 }
--- a/mock/src/tests.rs
+++ b/mock/src/tests.rs
@@ -6,7 +6,7 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use super::{Builder, Config, Event, Network, OurType, Peer, QuicP2p};
+use super::{Builder, Config, Event, EventSenders, Network, OurType, Peer, QuicP2p};
 use bytes::Bytes;
 use crossbeam_channel::{self as mpmc, Receiver, TryRecvError};
 use fxhash::FxHashSet;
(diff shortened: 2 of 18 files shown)


tests/quic_p2p.rs
```diff
@@ -14,16 +14,60 @@ fn wait_till_connected(ev_rx: mpmc::Receiver<Event>) -> Peer {
     panic!("Didn't receive the expected ConnectedTo event");
 }
```
dirvine: This seems duplicate to the above code in test_utils? Is this perhaps an indication of some code/tests being in the wrong place?

dirvine (commented):
Just small comment, possible code duplication. apart fomr that looks good to me. Nice work

tests/quic_p2p.rs
jeanphilippeD: Some are integration and other unit tests, and other examples.
The problem is integration tests or example can only use what is `pub` from the library code.

jeanphilippeD:
Rebase on latest master.

jeanphilippeD:
Remove unnecessary unused attribute.
Fix the mock implementation.

examples/common/mod.rs
```diff
@@ -16,3 +17,67 @@ pub enum Rpc {
...
+        let node_idx = sel.recv(&self.node_rx);
+        let selected_operation = sel.ready();
+
+        if selected_operation == client_idx {
+            self.client_rx.recv()
+        } else if selected_operation == node_idx {
+            self.node_rx.recv()
+        } else {
+            panic!("invalid operation");
+        }
```
madadam: I think this would be simpler using the [select](https://docs.rs/crossbeam-channel/0.4.0/crossbeam_channel/macro.select.html) macro.

examples/common/mod.rs
```diff
@@ -16,3 +17,67 @@ pub enum Rpc {
...
+
+        let event = if selected_operation == client_idx {
+            self.event_rx.client_rx.recv()
+        } else if selected_operation == node_idx {
+            self.event_rx.node_rx.recv()
+        } else {
+            return None;
+        };
+
+        event.ok()
```
madadam: Same here.

mock/src/tests.rs
```diff
@@ -28,6 +28,28 @@ macro_rules! assert_match {
...
+struct EventReceivers {
+    pub node_rx: Receiver<Event>,
+    pub client_rx: Receiver<Event>,
+}
+
+impl EventReceivers {
+    pub fn try_recv(&self) -> Result<Event, mpmc::TryRecvError> {
+        self.node_rx
+            .try_recv()
+            .or_else(|_| self.client_rx.try_recv())
```
madadam: I'd also use `select` here, otherwise the operation is biased towards  `node_rx`.

src/test_utils.rs
```diff
@@ -173,12 +173,78 @@ fn tmp_rand_dir() -> PathBuf {
...
+        let node_idx = sel.recv(&self.node_rx);
+        let selected_operation = sel.ready();
+
+        if selected_operation == client_idx {
+            self.client_rx.recv()
+        } else if selected_operation == node_idx {
+            self.node_rx.recv()
+        } else {
+            panic!("invalid operation");
+        }
```
madadam: Use `select` macro here too.

src/test_utils.rs
```diff
@@ -173,12 +173,78 @@ fn tmp_rand_dir() -> PathBuf {
...
+    }
+
+    pub fn try_recv(&self) -> Result<Event, mpmc::TryRecvError> {
+        self.node_rx
+            .try_recv()
+            .or_else(|_| self.client_rx.try_recv())
```
madadam: ...and here...

src/test_utils.rs
```diff
@@ -173,12 +173,78 @@ fn tmp_rand_dir() -> PathBuf {
...
+
+        let event = if selected_operation == client_idx {
+            self.event_rx.client_rx.recv()
+        } else if selected_operation == node_idx {
+            self.event_rx.node_rx.recv()
+        } else {
+            return None;
+        };
+
+        event.ok()
```
madadam: ...and also here.

tests/quic_p2p.rs
```diff
@@ -16,16 +16,60 @@ fn wait_till_connected(ev_rx: mpmc::Receiver<Event>) -> Peer {
...
+
+        let event = if selected_operation == client_idx {
+            self.event_rx.client_rx.recv()
+        } else if selected_operation == node_idx {
+            self.event_rx.node_rx.recv()
+        } else {
+            return None;
+        };
+
+        event.ok()
```
madadam: Same.

src/lib.rs
```diff
@@ -98,17 +98,36 @@ pub const DEFAULT_MAX_ALLOWED_MSG_SIZE: usize = 500 * 1024 * 1024; // 500MiB
...
+#[derive(Clone)]
+pub struct EventSenders {
+    /// The client event sender
+    pub client_tx: mpmc::Sender<Event>,
+    /// The node event sender
+    pub node_tx: mpmc::Sender<Event>,
+}
+
+impl EventSenders {
+    pub(crate) fn send(&self, event: Event) -> Result<(), mpmc::SendError<Event>> {
```
madadam: Adding this comment here but it applies generally to the whole design of this. There is an ergonomy issue in that the variants of the `Event` enum contain `Peer` which can be `Client` or `Node`, which means that even though you receive an event through say the `node` channel, you still need to at least nominally handle the `Client` case (using `unreachable!` for example). It would be better, API-wise, to have only `SocketAddr` instead of `Peer` in those `Event` variants and to use separate send functions: `send_client` and `send_node`.

madadam (commented):
Looks mostly good. Some code can be simplified using the `select` macro from crossbeam-channel and I left a suggestion to improve ergonomy.

tests/quic_p2p.rs
nbaksalyar: Maybe it's worth exposing `EventReceiever` as a helper struct then? As it's used in almost the same way in examples, integration tests, and unit tests, moving it to `test_utils` and making it public might be a good way to handle that.

src/lib.rs
```diff
@@ -98,17 +98,36 @@ pub const DEFAULT_MAX_ALLOWED_MSG_SIZE: usize = 500 * 1024 * 1024; // 500MiB
 /// before using a random port.
 pub const DEFAULT_PORT_TO_TRY: u16 = 443;
 
+/// Senders for node and client events
+#[derive(Clone)]
+pub struct EventSenders {
+    /// The client event sender
+    pub client_tx: mpmc::Sender<Event>,
+    /// The node event sender
+    pub node_tx: mpmc::Sender<Event>,
```
nbaksalyar: A general comment about naming these vars: I think it can be a bit confusing for the API user because `node_rx` / `client_rx` do not strictly relate to "nodes" only because I can use quic-p2p on the client side, and I may expect to receive e.g. a `BootstrappedTo` event - which in this case will be sent over the `node_tx` channel. So it's a bit counterintuitive that I'll need to use `node_rx` and _not_ `client_rx` for these events as a network client.

Probably a better name would be something like `outbound_tx` / `inbound_tx` or something like that?

src/lib.rs
jeanphilippeD: That would be fine with me, I think it is probably something that can be added by a follow up PR, as it is more of an additional change.

src/lib.rs
jeanphilippeD: > do not strictly relate to "nodes" only because I can use quic-p2p on the client side, and I may expect to receive e.g. a BootstrappedTo event 

My understanding if we cannot bootstrap on a client.

examples/common/mod.rs
jeanphilippeD: The select macro do not return the error. Also manual select is how we do it in vault because the source are separated. Apply to all of these comments.

src/lib.rs
jeanphilippeD: Raised as https://github.com/maidsafe/routing/issues/2053

mock/src/tests.rs
jeanphilippeD: Would need to use try_select. does not seem really necessary for a test function.

examples/common/mod.rs
madadam: > The select macro do not return the error.

I think it does?

```rust
select! {
    recv(self.client_rx) -> ev => /* ev is Result */,
    recv(self.node_rx) -> ev => /* ev is Result */,
}
```

> Also manual select is how we do it in vault because the source are separate

We do it like that because the set of channels to select over isn't statically known. But that's not the case here. `select!` would work just fine and make things simpler.

madadam: That said, if you don't want to do this due to lack of time or whatever then I'm OK leaving this to a followup PR.

mock/src/tests.rs
madadam: > Would need to use try_select

No, you would use `select!` with a `default` clause: 

```rust
select! {
    recv(self.node_rx) -> ev => ev,
    recv(self.client_rx) -> ev => ev,
    default => Err(TryRecvError::Empty)
}
```

madadam: Agree that this is probably not necessary though.

tests/quic_p2p.rs
jeanphilippeD: I don't think so, though this can always be updated later. But essentially we provide 2 channels so 2 different consumer can listen for the different events. The reason this struct make sense is in the existing tests we want to listen to both channels in the same place.

src/lib.rs
jeanphilippeD: Updated docs.

examples/common/mod.rs
jeanphilippeD: Correct, I'll raise an issue.

jeanphilippeD: Created issue https://github.com/maidsafe/quic-p2p/issues/96 and added it to the fleming routing project.

madadam (approved):

PR #97, opened 2020-02-13
Force the use of basic tokio scheduler

nbaksalyar:
If quic-p2p was included as a dependency in a repo with tokio dep listing a rt-threaded feature, it defaulted to use the threaded scheduler, resulting in a failure of finding the quic-p2p context (which is a thread local variable). This commit provides a fix for this problem by forcing the use of the basic, single-threaded tokio scheduler.

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -79,8 +79,11 @@ impl EventLoop {
                     }
                 };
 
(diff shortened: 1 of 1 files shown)


jeanphilippeD (approved):
Looks reasonable to me.

PR #106, opened 2020-05-14
feat(audit): add scheduled security audit scan

S-Coyle:
As per https://github.com/actions-rs/audit-check#scheduled-audit this will auto create a security issue if anything is found


S-Coyle:
Closing PR - seems I need to create the Action workflow via the actions tab if I want it to run against this PR

PR #111, opened 2020-05-21
Improve labelling of jobs in workflows

lionel-faber:
This PR also reduces the verbosity of the workflow files


S-Coyle (approved):
Nice, much more readable 👏

PR #132, opened 2020-08-27
Use single endpoint for each quic-p2p instance

lionel1704:
- all connections will be created using this end point


lionel1704:
Closing this since we are going in a different direction. Design wise.

PR #134, opened 2020-08-31
Allow multiple reads / writes in a stream

lionel-faber:


src/connections.rs
```diff
@@ -43,76 +39,71 @@ impl Connection {
...
+
+    /// Send message to peer and await for a reponse.
+    pub async fn send(&self, msg: Bytes) -> Result<ConnectionStreams> {
+        let mut connection_streams = self.create_connection_streams().await?;
+        connection_streams.send(msg).await?;
+        Ok(connection_streams)
     }
 
     /// Send message to peer using a bi-directional stream without awaiting for a reponse.
     pub async fn send_only(&self, msg: Bytes) -> Result<()> {
```
bochaco, line 62: I'm wondering if we should just remove this function now, if the user wants to only send on a bi-stream then `send` can be used and the result returned can be simply dropped by the user obtaining the same result?

src/connections.rs
```diff
@@ -43,76 +39,71 @@ impl Connection {
         self.quic_conn.remote_address()
     }
 
-    /// Send message to peer and await for a reponse.
-    pub async fn send(&mut self, msg: Bytes) -> Result<Bytes> {
+    /// Get connection streams for reading/writing
+    pub async fn create_connection_streams(&self) -> Result<ConnectionStreams> {
```
bochaco, line 43: I was thinking we may want to return two separate objects here so the user can spawn tasks to send and receive on the bi-stream from separate tasks/threads.

lionel1704: Yes that makes sense :+1:

src/connections.rs
joshuef: That'd still give you unnecessary wait time, no?

For `send` to work, you'd have to `await` the response, regardless of if you're going to use it?

lionel1704: That's changed now. send will return the streams so you can do `recv.read()` if you want the response or just drop them if you don't expect a response.

src/connections.rs
```diff
@@ -240,38 +191,70 @@ impl IncomingMessages {
...
-                }
-            }
-            Err(err) => {
-                warn!("Failed reading message's bytes: {}", err);
-                None
-            }
+/// Read the message's bytes which size is capped
+pub async fn read_bytes(recv: &mut quinn::RecvStream, _max_msg_size: usize) -> Result<Bytes> {
+    let mut data_len: [u8; 1] = [0; 1];
+    recv.read_exact(&mut data_len).await?;
```
Yoga07, line 205: Why do we call `read_exact` twice(here with a slice and below with a vec)? Recon it's for the flag?

src/connections.rs
```diff
@@ -240,38 +191,70 @@ impl IncomingMessages {
...
+        WireMsg::UserMsg(msg_bytes) => Ok(Bytes::copy_from_slice(&msg_bytes)),
+        WireMsg::EndpointEchoReq | WireMsg::EndpointEchoResp(_) => {
+            // TODO: handle the echo request/response message
+            unimplemented!("echo message type not supported yet");
         }
     }
 }
 
+/// Helper to send bytes to peer using the provided stream.
+pub async fn send_msg(
```
Yoga07, line 219: Should this be `pub`?

Yoga07 (commented):
Looks neat to me! Just minor ones for clarification.

src/connections.rs
lionel1704: I made it public temporarily. In the latest commits I've made it private again.

src/connections.rs
bochaco: This is to obtain the length of the message. 

Although, @lionel1704, I think this is where we should use the `WireMsg`, that struct is already being used for prefixing the message bytes with information which is handled only by the quic-p2p layer, same as the length we need now, so we should add a `msg_length` field (fixed size) to `WireMsg` and do a first read from the stream of length == size_of(<`WireMsg` struct>). Once you read that, you parse it with `WireMsg::from_raw(data)` and from there you obtain the info about message type and message length. Then if what you have is `WireMsg::UserMsg({ msg_bytes, msh_length })` you can read a second time from the stream with a slice/Vec of size == msg_length.

bochaco (changes requested):

src/connections.rs
lionel1704: >  do a first read from the stream of length == size_of(<WireMsg struct>)

But, this length isn't fixed. So we wouldn't really know how many bytes to read?

bochaco: You are right, the msg bytes should be removed from it, so the WireStruct should have only the length field.

joshuef (approved):

PR #158, opened 2020-10-01
Async UPnP and Echo service

lionel-faber:


bochaco (changes requested):
I think we should consider if it's possible to remove the feature `upnp` as it make the code more unreadable and perhaps it's over killing, perhaps we just need a function which is specific for using upnp for discovering the public address, and the consumer of the api decides to call it if that's what it needs.

src/connections.rs
```diff
@@ -225,8 +230,17 @@ impl IncomingMessages {
...
                 None
             }
-            Some(Ok((send, mut recv))) => match read_bytes(&mut recv).await {
-                Ok(bytes) => Some((bytes, send, recv)),
+            Some(Ok((mut send, mut recv))) => match read_bytes(&mut recv).await {
+                Ok(WireMsg::UserMsg(bytes)) => Some((bytes, send, recv)),
+                Ok(WireMsg::EndpointEchoReq) => {
+                    let message = WireMsg::EndpointEchoResp(peer_addr);
+                    message.write_to_stream(&mut send).await.ok()?;
+                    None
```
bochaco, line 238: I think we shouldn't return `None` in this case as that would mean for the user that there won't be no more messages arriving in the stream, and therefore the user will end the loop. We should send the echo response, but then go back and read the next message by continuing the loop, so if next message is for the user it will receive it as thee next message without noticing there was an echo response sent in beforehand.

src/connections.rs
```diff
@@ -225,8 +230,17 @@ impl IncomingMessages {
...
+                }
+                Ok(msg) => {
+                    error!("Unexpected message type: {:?}", msg);
+                    None
```
bochaco, line 242: Same as above comment, we should log the error but we shouldn't return `None`, instead read next messages by continuing the loop.

lionel1704:
These changes are incorporated in #162

PR #162, opened 2020-10-27
Add --clean and --fresh configurations

lionel-faber:
This would help prevent the use of outdated configurations


S-Coyle (dismissed):
Did some testing with the new commands on macOS and Ubuntu - all behaving as expected

joshuef (dismissed):

Cargo.toml
```diff
@@ -44,19 +45,14 @@ webpki = "~0.21.3"
...
-  optional = true
 
   [dependencies.rustls]
   version = "~0.17.0"
   features = [ "dangerous_configuration" ]
 
 [dev-dependencies]
-clap = "2.33.1"
-tracing = "~0.1.15"
+async-std = "1.6.5"
```
madadam, line 54: Why do you need `async-std`? Can't you just use tokio? Note: I'm not saying tokio is better than async-std (or vice-versa), just that it seems unnecessary to bring in another runtime when we already have tokio.

examples/echo_service.rs
```diff
@@ -0,0 +1,90 @@
...
+        }
+    } else {
+        println!("Echo service complete");
+        let node_addr = bootstrap_nodes[0];
+        let connection = endpoint.connect_to(&node_addr).await?;
+        let (mut send, mut recv) = connection.open_bi_stream().await?;
+        loop {
+            println!("Enter message:");
+            let mut input = String::new();
+            std::io::stdin().read_line(&mut input)?;
```
madadam, line 74: Using blocking functions inside `async` is not recommended. Why not use the `read_from_stdin` function defined below?

examples/echo_service.rs
```diff
@@ -0,0 +1,90 @@
...
+                std::str::from_utf8(&bytes[..])
+                    .map_err(|_| Error::Unexpected("Irregular String format".to_string()))?
+            );
+        }
+    }
+}
+
+async fn read_from_stdin() -> String {
+    let mut input = String::new();
+    io::stdin().read_line(&mut input).await.unwrap_or(0);
```
madadam, line 88: tokio has [io::stdin](https://docs.rs/tokio/0.2.22/tokio/io/fn.stdin.html) as well, so no need to use `async-std`. Just need to also bring [AsyncReadExt](https://docs.rs/tokio/0.2.22/tokio/io/trait.AsyncReadExt.html) into scope so you can use `read`, `read_line`, and so.

examples/echo_service.rs
lionel1704: The first message needs to be send by the peer that initializes the connection. That's why this is synchronous.

madadam (approved):

PR #172, opened 2020-11-16
Connection pool

madadam:
Connections are now stored in a pool and reused. See the doc comment for `Endpoint::connect_to` for more details about how this works. 

This is API breaking:

- `Endpoint::connect_to` return type changed from `Result<Connection>` to `Result<(Connection, Option<IncomingMessages>)>`.
- Similarly, `QuicP2p::bootstrap` return type changed from `Result<(Endpoint, Connection)>` to `Result<(Endpoint, Connection, IncomingMessages)>`
- `Connection` no longer automatically closes the underlying `quinn::Connection` on drop. It is only closed when the `quinn::Connection` and all its associated streams get dropped. 
- `Connection::send` renamed to `Connection::send_bi` and `Connection::open_bi_stream` renamed to `Connection::open_bi` - for consistency
- `Endpoint::listen` no longer returns `Result`

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -19,7 +19,6 @@ use std::{
     collections::HashSet, fmt, fs, io, net::IpAddr, net::SocketAddr, path::PathBuf, str::FromStr,
 };
 use structopt::StructOpt;
-use unwrap::unwrap;
 
 /// QuicP2p configurations
 #[derive(Clone, Debug, Default, Serialize, Deserialize, Eq, PartialEq, StructOpt)]
@@ -135,12 +134,10 @@ impl SerialisableCertificate {
 
 impl Default for SerialisableCertificate {
     fn default() -> Self {
-        let cert = unwrap!(rcgen::generate_simple_self_signed(vec![
-            "MaidSAFE.net".to_string()
-        ]));
+        let cert = rcgen::generate_simple_self_signed(vec!["MaidSAFE.net".to_string()]).unwrap();
 
         Self {
-            cert_der: unwrap!(cert.serialize_der()).into(),
+            cert_der: cert.serialize_der().unwrap().into(),
             key_der: cert.serialize_private_key_der().into(),
         }
     }
@@ -160,7 +157,7 @@ impl FromStr for SerialisableCertificate {
 impl ToString for SerialisableCertificate {
     /// Convert `SerialisableCertificate` into a base64-encoded string.
     fn to_string(&self) -> String {
-        let cert = unwrap!(bincode::serialize(&(&self.cert_der, &self.key_der)));
+        let cert = bincode::serialize(&(&self.cert_der, &self.key_der)).unwrap();
         base64::encode(&cert)
     }
 }
@@ -195,12 +192,12 @@ mod tests {
     #[test]
     fn config_create_read_and_write() {
         let dir = test_dirs();
-        let config_path = unwrap!(config_path(Some(&dir)));
+        let config_path = config_path(Some(&dir)).unwrap();
 
         assert!(utils::read_from_disk::<Config>(&config_path).is_err());
 
-        let cfg = unwrap!(Config::read_or_construct_default(Some(&dir)));
-        let read_cfg = unwrap!(utils::read_from_disk(&config_path));
+        let cfg = Config::read_or_construct_default(Some(&dir)).unwrap();
+        let read_cfg = utils::read_from_disk(&config_path).unwrap();
 
         assert_eq!(cfg, read_cfg);
     }
--- a/src/connection_deduplicator.rs
+++ b/src/connection_deduplicator.rs
@@ -1,0 +1,86 @@
+// Copyright 2020 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under the MIT license <LICENSE-MIT
+// http://opensource.org/licenses/MIT> or the Modified BSD license <LICENSE-BSD
+// https://opensource.org/licenses/BSD-3-Clause>, at your option. This file may not be copied,
+// modified, or distributed except according to those terms. Please review the Licences for the
+// specific language governing permissions and limitations relating to use of the SAFE Network
+// Software.
+
+use super::connections::Connection;
+use err_derive::Error;
+use std::{
+    collections::{hash_map::Entry, HashMap},
+    net::SocketAddr,
+};
+use tokio::sync::{broadcast, Mutex};
+
+type Result<T = Connection, E = Error> = std::result::Result<T, E>;
+
+// Deduplicate multiple concurrent connect attempts to the same peer - they all will yield the
+// same `Connection` instead of opening a separate connection each.
+pub(crate) struct ConnectionDeduplicator {
+    map: Mutex<HashMap<SocketAddr, broadcast::Sender<Result>>>,
+}
+
+impl ConnectionDeduplicator {
+    pub fn new() -> Self {
+        Self {
+            map: Mutex::new(HashMap::new()),
+        }
+    }
+
+    // Query if there already is a connect attempt to the given address.
+    // If this is the first connect attempt, it returns `None` and we should proceed with
+    // establishing the connection and then call `complete` with the result.
+    // For any subsequent connect attempt this returns `Some` with the result of that attempt.
+    pub async fn query(&self, addr: &SocketAddr) -> Option<Result> {
+        let mut rx = match self.map.lock().await.entry(*addr) {
+            Entry::Occupied(entry) => entry.get().subscribe(),
+            Entry::Vacant(entry) => {
+                let (tx, _) = broadcast::channel(1);
+                let _ = entry.insert(tx);
+                return None;
+            }
+        };
+
+        if let Ok(result) = rx.recv().await {
+            Some(result)
+        } else {
+            // NOTE: this branch cannot realistically happen, because we never drop the `Sender`
+            // without sending through it first and we also take it out of the map before doing so
+            // which means it's not possible to create more subscription on it after the send.
+            // This, however, is not statically provable so we still need to nominally handle this
+            // branch. We return `LocallyClosed` which seems like it would be the right error to
+            // return if this situation was actually possible (it isn't) as it would signal the
+            // caller to either abandon the connect or try to repeat it.
+            Some(Err(quinn::ConnectionError::LocallyClosed.into()))
+        }
+    }
+
+    // Signal completion of a connect attempt. This causes all the pending `query` calls for the
+    // same `addr` to return `result`.
+    pub async fn complete(&self, addr: &SocketAddr, result: Result) {
+        let tx = self.map.lock().await.remove(addr);
+        if let Some(tx) = tx {
+            let _ = tx.send(result);
+        }
+    }
+}
+
+#[derive(Clone, Error, Debug)]
+pub(crate) enum Error {
+    #[error(display = "Connect error")]
+    Connect(#[source] quinn::ConnectError),
+    #[error(display = "Connection error")]
+    Connection(#[source] quinn::ConnectionError),
+}
+
+impl From<Error> for crate::error::Error {
+    fn from(src: Error) -> Self {
+        match src {
+            Error::Connect(source) => Self::Connect(source),
+            Error::Connection(source) => Self::Connection(source),
+        }
+    }
+}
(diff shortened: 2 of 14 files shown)


src/config.rs
```diff
@@ -135,12 +134,10 @@ impl SerialisableCertificate {
...
 impl Default for SerialisableCertificate {
     fn default() -> Self {
-        let cert = unwrap!(rcgen::generate_simple_self_signed(vec![
-            "MaidSAFE.net".to_string()
-        ]));
+        let cert = rcgen::generate_simple_self_signed(vec!["MaidSAFE.net".to_string()]).unwrap();
 
         Self {
-            cert_der: unwrap!(cert.serialize_der()).into(),
+            cert_der: cert.serialize_der().unwrap().into(),
```
bochaco, line 140: I guess we want to get rid of unwraps, not sure what's the best option since default() cannot return Result...

src/connection_deduplicator.rs
```diff
@@ -0,0 +1,86 @@
...
+
+use super::connections::Connection;
+use err_derive::Error;
+use std::{
+    collections::{hash_map::Entry, HashMap},
+    net::SocketAddr,
+};
+use tokio::sync::{broadcast, Mutex};
+
+type Result<T = Connection, E = Error> = std::result::Result<T, E>;
```
bochaco, line 18: Is this a common practice? I find it weird as when reading the code i cannot see what a function returns...? shouldn't be better to keep `T` to be set by insstantiator ?

src/connection_pool.rs
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    pub fn insert(&self, addr: SocketAddr, conn: quinn::Connection) -> ConnectionHandle {
+        // NOTE: this will panic if some other thread panicked while holding this lock. That's
+        // probably the correct thing to do as it avoids working with potentially invalid state
+        // (see the rust docs about mutex poisoning for more info).
+        // That said, it's highly unlikely this will ever panic as none of the operations we
+        // perform anywhere while holding the lock can panic.
+        // This comment applies to all the other instances of `unwrap` in this module.
+        let mut store = self.store.lock().unwrap();
```
bochaco, line 37: Shouldn't we return `Result` still? so if one day someone introduces a bug in some other part of the code which opens up the possibility of a panic then the user app gets an error and can handle it or report it properly

src/connections.rs
```diff
@@ -19,19 +20,18 @@ use std::{net::SocketAddr, sync::Arc};
 use tokio::select;
 
 /// Connection instance to a node which can be used to send messages to it
+#[derive(Clone)]
 pub struct Connection {
     quic_conn: quinn::Connection,
```
bochaco, line 25: How about keeping this within the `ConnectionHandle` instance?, its name suggested me up front I'd be using this handle to make use of the connection and when dropped it internally handles the removal from the pool and disconnection accordingly.

src/connections.rs
```diff
@@ -19,19 +20,18 @@ use std::{net::SocketAddr, sync::Arc};
...
-}
-
-impl Drop for Connection {
-    fn drop(&mut self) {
-        self.close();
-    }
+    pool_handle: ConnectionHandle,
```
bochaco, line 26: Shouldn't we call it `conn_handle` instead as we are not really interested in the pool here but just the connection we got the handle for?

src/connections.rs
```diff
@@ -141,43 +160,43 @@ impl IncomingConnections {
 
 /// Stream of incoming QUIC messages
 pub struct IncomingMessages {
-    peer_addr: SocketAddr,
     uni_streams: quinn::IncomingUniStreams,
     bi_streams: quinn::IncomingBiStreams,
+    pool_handle: ConnectionHandle,
```
bochaco, line 165: Same Q about the name of the var as above...?

src/tests/common.rs
```diff
@@ -140,12 +140,18 @@ async fn uni_directional_streams() -> Result<()> {
         assert_eq!(src, peer2_addr);
         src
     } else {
-        return Err(Error::UnexpectedMessageType);
+        panic!("Expected a unidirectional stream")
```
bochaco, line 143: Any reason we are going back to panic here instead of returning Result?

src/wire_msg.rs
```diff
@@ -63,7 +62,7 @@ impl WireMsg {
         let (msg_bytes, msg_flag) = match self {
             WireMsg::UserMsg(ref m) => (m.clone(), USER_MSG_FLAG),
             _ => (
-                From::from(unwrap!(bincode::serialize(&self))),
+                From::from(bincode::serialize(&self).unwrap()),
```
bochaco, line 65: This function returns a Result so I guess we can return an Err here easily ?

bochaco (changes requested):

src/connection_pool.rs
dirvine: Yes, let's get rid of unwrap in all of our code with no debates about it. Even locks can poison and if they do we can still get the data from the thread if we want and so on, so possible to handle or just return result as well. We have a policy, no unwraps in production so let's Engineer our code to not have them.

src/config.rs
dirvine: Perhaps we cannot have default in this case? Should be possible to Engineer this out AFAIK.

madadam: In this PR I only removed the `unwrap!` macro by replacing it with regular `unwrap`. I know we want to remove those as well, but maybe best to take that in a separate PR? I though @lionel1704 was on it?

src/connection_deduplicator.rs
madadam: This is just to reduce typing. It is not exposed outside of this module. 

> Is this a common practice

Well, I can think of at least one precedent: https://doc.rust-lang.org/std/fmt/type.Result.html

src/connection_pool.rs
madadam: I think calling `unwrap` on the result of `Mutex::lock` is one of the few legitimate uses of `unwrap` in production code. The reason is that `lock` returns error only if the mutex is poisoned which means some other thread panicked while holding the lock. So if this `unwrap` panics, that means there already is a panic somewhere else in our code which means our goal of never panicking in production has already failed :smiling_imp: Also, if you look at the other places where we lock this mutex and the things we do while it's being locked then you'll see that nothing there can actually panic. That means that this `unwrap` never panics either. So propagating this error upwards would just make the API annoying to use for no good reason.

src/connections.rs
madadam: Hm, the purpose of `ConnectionHandle` is just to remove the connection from the pool. So maybe it should be called something else? Something like `ConnectionPoolRemover` or so?

> when dropped it internally handles the removal from the pool 

This would defy the purpose of connection pool, because then you would have to stash the connection handle somewhere, otherwise the connection would be closed. The idea is to now need any stash. Just fetch the connection by calling `Endpoint::connect_to` and it either returns the connection from the pool or opens a new one, but the caller doesn't need to worry about that.

src/connections.rs
madadam: Well it's purpose is to remove the connection from the pool, so having `pool` in the name makes sense in my opinion. But as said in the previous comment, maybe the name isn't clear?

src/tests/common.rs
madadam: We are not going back, but I think in this case `panic` is more appropriate because this is actually a form of assertion. It's just not expressed with the `assert` macro because there is no macro that asserts that something matches a pattern (there is `assert_matches` crate which we use in routing but I wasn't sure we want to add it here as well).

src/wire_msg.rs
madadam: https://github.com/maidsafe/qp2p/pull/172#discussion_r526156207

src/connection_pool.rs
dirvine: There will only be a panic elsewhere if there is code that panics in there. So allowing panic is how that happens. 
If the lock panicked we can still get the data by error handling but should probably pass back an error we cannot handle or are unlikely to. 
Safe to say code can panic in places where we `unwrap` and cannot where we don't. 
We will need to check dependencies at some stage to make sure they don't cause panics anyway as that is another wy to poison a thread we think can never be poisoned, it's worse though if we have panics in our code as well. 
IMO it's so much simpler to audit code that does not have the world panic/unwrap in it as then I need to start looking at the mem internals or thread poisoning or some other thing and the logic gets lost. 
So I would say panic nowhere and we stand a chance. 
Atm we cannot tell if some dependent crate will panic as we need to audit them, until that time these threads can poison. Even then we can still handle it with restructuring the error and deciding what to do there.

src/tests/common.rs
dirvine: >(there is assert_matches crate which we use in routing but I wasn't sure we want to add it here as well).
I would perhaps @madadam  if it allows us to get rid of a panic as assert is more obvious in tests IMO.

src/wire_msg.rs
dirvine: Not getting this, if that panics it kills a long-running node. WE defo don't want to do that.

src/connection_pool.rs
bochaco: > Also, if you look at the other places where we lock this mutex and the things we do while it's being locked then you'll see that nothing there can actually panic. That means that this unwrap never panics either.
This is what I was trying to address in my previous comment, in the future someone may add/change code in those places which start panicking, specially if they see we use panic in some places like here, so by having any unwrap we implicitly say "we allow our code to panic sometimes" then new devs will start adding unwraps which will break the invariant you are referring to.

src/connections.rs
bochaco: It's not a big deal, but then why the struct is named ConnectionHandle, whatever that struct is called you are not getting a handle to the pool, but a handle to somethings that knows how to remove the connection from the pool, so that's what I mean.

src/connections.rs
bochaco: right, i am probably misunderstanding, I thought the pool has a clone of the quinn::Connection thus we could reach/use it through the ConnectionHandler, e.g. ConnectionHandler::get_conn() which gives a clone of quinn::Connection ...??....

src/connection_pool.rs
madadam: OK guys, will change this to use `PoisonError::into_inner` just to not waste time on this even though I disagree with you on this one.

src/connections.rs
madadam: No, `ConnectionHandle` is only for removing. I'm going to rename it to make it more clear.

src/wire_msg.rs
madadam: I don't disagree with that. I'm just saying it's not the point of this PR and that there already is a work by @lionel1704 to address this so perhaps it should be done there too?

src/tests/common.rs
madadam: OK, will use `assert_matches`.

src/config.rs
lionel1704: I have raised madadam/quic-p2p#1 to this branch already. I can update the PR if needed.

src/wire_msg.rs
dirvine: IF Lionel has this covered then cool, otherwise let's change here. @lionel1704  can you confirm this one.

lionel1704: Yes that work is already done on top of this branch. Once this is merged I can raise a PR with those changes.

src/connection_pool.rs
dirvine: Also @madadam  Can you put a TODO here 
//TODO Check data plus invariants here!!

madadam: How about doing that thing now instead of TODO it? But then I'm not sure what you mean by "check data plus invariants". Could you elaborate?

dirvine: Yea no worries, not in all cases would this be an error we cannot "fix" (this is why we don't want panic by default). 
So in these cases we can look at what was in the thread behind the guard and if there is data we need, then we can still access it, if the invariants all still hold then we can continue (even if we need to do some other work), otherwise we would pass this back and see if anyone else can handle it. If not it's all the way back to main and fails (iff invariants are broken)

bochaco (approved):

PR #179, opened 2020-11-19
Remove all unwrap, expects and panics

lionel-faber:


src/peer_config.rs
```diff
@@ -26,18 +26,19 @@ pub const DEFAULT_KEEP_ALIVE_INTERVAL_MSEC: u32 = 10_000; // 10secs
...
     idle_timeout_msec: u64,
     keep_alive_interval_msec: u32,
-) -> quinn::ClientConfig {
+) -> Result<quinn::ClientConfig> {
     let mut cfg = quinn::ClientConfigBuilder::default().build();
-    let crypto_cfg =
-        Arc::get_mut(&mut cfg.crypto).expect("the crypto config should not be shared yet");
+    let crypto_cfg = Arc::get_mut(&mut cfg.crypto).ok_or_else(|| {
+        Error::Unexpected("The Crypto config should not be shared yet".to_string())
+    })?;
```
madadam, line 33: I think in this case using `expect` was actually the right thing to do. This is because this would only return `None` if there was a bug in the code (in this case the bug would be in quinn). I don't think it's reasonable to propagate bugs into errors. This is what panicking is actually for. 

Arguably the problem here is on the quinn side - they should offer an infallible way to construct the client config. It's a bad API design on their side. Might be worth raising an issue there. But until that's addressed, I'd stick with `expect` here.

madadam: Actually, it seems it's possible to make this infallible by ourselves. We just need to use `quinn::ClientConfig` and `rustls::ClientConfig` directly. Something like:
```rust
let mut crypto = rustls::ClientConfig::default();
crypto.dangerous().set_certificate_verifier(SkipServerVerification::new());
let transport = new_transport_cfg(idle_timeout_msec, keep_alive_interval_msec);
quinn::ClientConfig {
    transport: Arc::new(transport),
    crypto: Arc::new(crypto),
}
```
This might be the best way to do this.

maqi (commented):
LGTM
just wodering, is this (https://github.com/lionel1704/quic-p2p/blob/no-unwrap/src/bootstrap_cache.rs#L144) still required ?

Also, just to confirm the two panics left in the `example\echo_service.rs` is ok, right?

lionel1704:
I can quickly address those.

maqi (approved):

PR #189, opened 2020-12-13
Execute GHAs only in maidsafe repository

Thierry61:
Forking Maidsafe repositories is a pain because they have GHA workflows that don’t work from other repositories du to secrets used in them. This generates errors and many mails with each push commands.

This PR, defined on qp2p as an example, solves this problem by executing actions only within Maidsafe repository but not in forked ones.

The solution is simple, add the following line to control execution of each action: `if: ${{ github.repository_owner == 'maidsafe' }}`.

Additionally, the bad indentation of tag_release.yml has been corrected.


dirvine (dismissed):

Thierry61:
As you have noticed, latest version of clippy doesn’t pass `field_reassign_with_default` requirement in src/peer_config.rs.
Suggested correction is the following:
```diff
diff --git a/src/peer_config.rs b/src/peer_config.rs
index d10a2e6..982e37f 100644
--- a/src/peer_config.rs
+++ b/src/peer_config.rs
@@ -46,11 +46,13 @@ pub fn new_our_cfg(
     our_key: quinn::PrivateKey,
 ) -> Result<quinn::ServerConfig> {
     let mut our_cfg_builder = {
-        let mut our_cfg = quinn::ServerConfig::default();
-        our_cfg.transport = Arc::new(new_transport_cfg(
-            idle_timeout_msec,
-            keep_alive_interval_msec,
-        ));
+        let our_cfg = quinn::generic::ServerConfig {
+            transport: Arc::new(new_transport_cfg(
+                idle_timeout_msec,
+                keep_alive_interval_msec,
+            )),
+            ..Default::default()
+        };

         quinn::ServerConfigBuilder::new(our_cfg)
     };
```
Unfortunately this doesn’t compile because `token_key`, `use_stateless_retry`, `retry_token_lifetime`, `accept_buffer`, `migration` fields are `pub(crate)` (in struct `quinn::generic::ServerConfig`). To make it work they should be changed to `pub`.

Someone from Maidsafe should make a PR to quinn people to change this, unless you find a better solution.

S-Coyle (approved):

PR #192, opened 2020-12-30
feature(default ip) Use localhost as IP.

dirvine:
Do not bind ever to 0.0.0.0

### DIFF
--- a/examples/echo_service.rs
+++ b/examples/echo_service.rs
@@ -1,7 +1,7 @@
 use anyhow::{anyhow, bail, Error, Result};
 use bytes::Bytes;
 use qp2p::{Config, Message, QuicP2p};
-use std::env;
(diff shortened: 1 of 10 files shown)


examples/echo_service.rs
```diff
@@ -23,7 +26,7 @@ async fn main() -> Result<(), Error> {
 
     let qp2p = QuicP2p::with_config(
         Some(Config {
-            ip: None,
+            ip: Some(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1))),
```
lionel1704, line 29: Using localhost here might beat the purpose of this example. The primary goal of this example is to test the echo service using which the public IP of the connecting node is determined. 

The ideal fix for this would be to error out when the IP is not specified when calling the example with `create`. 
I can raise a PR with the fix for this and also document this in comments if that's okay?

PR #193, opened 2020-12-31
chore(tokio): upgrade tokio to v1.2.0, quinn to v0.7.0 and rustls to v0.19.0

bochaco:


lionel1704:
Looking good! We just need to wait for the quinn release now. I'll run some e2e tests with this branch of qp2p and the other tokio v1 branches to verify that everything runs smoothly.

dan-da:
How does this compare with #235 and #214?  Have those been evaluated?

lionel1704:
I don't think re-exporting tokio(#235) is desirable. 
#214 is very similar to this PR. It's pretty much the same.

lionel-faber (approved):

PR #196, opened 2021-01-14
Add cargo deny CI check

S-Coyle:


S-Coyle:
Merging as issue seen in unused dep check is unrelated to this PR - also being seen in routing.

PR #206, opened 2021-02-01
New and improved qp2p API

lionel-faber:
This PR introduces a new and improved simple qp2p API. This also implements connection and message handling within the qp2p crate itself so that the user of qp2p does not need to manage them.


src/endpoint.rs
```diff
@@ -240,28 +339,46 @@ impl Endpoint {
...
+            Some(Connection::new(conn, guard))
+        } else {
+            None
+        }
+    }
+
+    /// Sends a message to a peer. This will attempt to re-use any existing connections
+    /// with the said peer. If a connection doesn't exist already, a new connection will be created.
+    pub async fn send_message(&self, msg: Bytes, dest: &SocketAddr) -> Result<()> {
+        self.connect_to(dest).await?;
```
madadam, line 370: Why not return the connection directly from `connect_to`? Then we wouldn't have to look it up again using `get_connection` here.

src/endpoint.rs
```diff
@@ -297,4 +417,20 @@ impl Endpoint {
...
+
+    /// Returns the socket address of a peer that has connected to us
+    pub async fn next_incoming_connection(&mut self) -> Option<SocketAddr> {
+        self.connection_queue.lock().await.pop_front()
+    }
+
+    /// Returns a message received from another peer along with the source address
+    pub async fn next_incoming_message(&mut self) -> Option<(SocketAddr, Bytes)> {
+        self.message_queue.lock().await.pop_front()
+    }
```
madadam, line 429: First of all, you don't need `&mut self` because the message queue is behind a `Mutex`. But more serious issue is that this seems like it forces us to busy-loop for incoming messages, wasting CPU cycles. Wonder why do we need the message queue at all. Why not just `recv` on the message channel receiver instead? This same point applies to `next_incoming_connection` and `next_disconnected_peer` too.

src/connections.rs
```diff
@@ -332,16 +176,178 @@ impl std::fmt::Debug for SendStream {
...
+                next_bi.map(|(bytes, send, recv)| Message::BiStream {
+                    bytes,
+                    src,
+                    send: SendStream::new(send),
+                    recv: RecvStream::new(recv)
+                }),
+            };
+            if let Some(message) = message {
+                message_tx
+                    .send((src, message.get_message_data()))
```
madadam, line 265: Why even create the `Message` enum when only the data is used here?

Cargo.toml
```diff
@@ -21,6 +21,7 @@ serde_json = "1.0.59"
 structopt = "~0.3.15"
 thiserror = "1.0.23"
 webpki = "~0.21.3"
+flexi_logger = "~0.16.1"
```
madadam, line 24: What's the reason of moving this from dev-dependencies to here?

src/endpoint.rs
```diff
@@ -149,11 +241,22 @@ impl Endpoint {
...
+            self.public_addr = Some(socket_addr);
             Ok(socket_addr)
+        })
+    }
+
+    /// Disconnects from a peer
+    pub fn disconnect_from(&mut self, peer_addr: &SocketAddr) -> Result<()> {
+        if let Some((conn, remover)) = self.connection_pool.remove(peer_addr) {
+            conn.close(0u8.into(), b"User Disconnected");
+            remover.remove();
```
madadam, line 255: You already removed the connection from the pool by calling `ConnectionPool::remove` above, so there is no need to remove it via the remover too.

src/connection_pool.rs
```diff
@@ -42,6 +42,35 @@ impl ConnectionPool {
...
+
+        // Efficiently fetch the first entry whose key is equal to `key` and check if it exists
+        store
+            .map
+            .range_mut(Key::min(*addr)..=Key::max(*addr))
+            .next()
+            .is_some()
+    }
+
+    pub fn remove(&mut self, addr: &SocketAddr) -> Option<(quinn::Connection, ConnectionRemover)> {
```
madadam, line 56: There is no point in returning the `Remover` here. The remover is a handle to remove the connection from the pool. But you are doing that manually in this function, so the remover has no purpose (calling `remove` on it would have no effect).

madadam: Also, this removes only one connection with that address, but in some cases (*) there can be more than one. Wonder if it wouldn't make more sense to remove all of them instead?

*) if two peers open a connection to each other around the same time, they can end up having two connections to the same peer in the pool (one incoming and one outgoing).

src/endpoint.rs
```diff
@@ -149,11 +241,22 @@ impl Endpoint {
...
         } else {
-            Err(Error::NoEchoServiceResponse)
+            log::info!("We are not connected to this peer. Doing nothing");
```
bochaco, line 257: Perhaps the message should be more informative about what it was attempting to do and that it's not doing it since there is not connection, otherwise when reading the logs you cannot understand what is it that "it's not doing".

lionel1704: Of course :) I was saving log clean-up for the end :smile:

src/endpoint.rs
lionel1704: The connection object is not a part of the public API. It is held internally and handled within this crate.

src/connections.rs
lionel1704: Yes. I'm going to remove the Message enum completely.

src/endpoint.rs
```diff
@@ -250,28 +349,54 @@ impl Endpoint {
...
+    ) -> Result<(SendStream, RecvStream)> {
+        self.connect_to(peer_addr).await?;
+        let connection = self
+            .get_connection(peer_addr)
+            .ok_or(Error::MissingConnection)?;
+        connection.open_bi().await
+    }
+
+    /// Sends a message to a peer. This will attempt to re-use any existing connections
+    /// with the said peer. If a connection doesn't exist already, a new connection will be created.
```
madadam, line 390: > If a connection doesn't exist already, a new connection will be created

This doesn't seem to match the implementation.

joshuef (approved):

PR #213, opened 2021-02-11
Allow compilation on 32 bit architectures

nashley:
Don't assume that usize is equivalant to a u64

### DIFF
--- a/src/wire_msg.rs
+++ b/src/wire_msg.rs
@@ -14,6 +14,7 @@ use crate::{
 use bytes::Bytes;
 use log::trace;
 use serde::{Deserialize, Serialize};
+use std::convert::TryFrom;
 use std::{fmt, net::SocketAddr};
 
 const MSG_HEADER_LEN: usize = 9;
@@ -25,7 +26,7 @@ pub enum WireMsg {
     EndpointEchoReq,
     EndpointEchoResp(SocketAddr),
     EndpointVerificationReq(SocketAddr),
-    EndpointVerficationResp(bool),
+    EndpointVerificationResp(bool),
     UserMsg(Bytes),
 }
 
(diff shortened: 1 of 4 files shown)


nashley:
I think I followed the [contribution guidelines](https://github.com/maidsafe/QA/blob/master/CONTRIBUTING.md), but please let me know if this is missing anything!

nashley:
https://github.com/rust-lang/rust/issues/70460 would allow this to be done more cleanly than only allowing compilation on platforms where we know the `try_from` won't fail.

src/wire_msg.rs
```diff
@@ -41,7 +42,12 @@ impl WireMsg {
 
         let msg_header = MsgHeader::from_bytes(header_bytes);
         log::debug!("reading data of length: {}", msg_header.data_len());
-        let mut data: Vec<u8> = vec![0; msg_header.data_len()];
+        // https://github.com/rust-lang/rust/issues/70460 for work on a cleaner alternative:
+        #[cfg(not(any(target_pointer_width = "32", target_pointer_width = "64")))]
+        {
+            compile_error!("You need an architecture capable of addressing 32 bit pointers");
+        }
+        let mut data: Vec<u8> = vec![0; usize::try_from(msg_header.data_len()).unwrap()];
```
lionel1704, line 50: We try to avoid using `.unwrap()` wherever we can to prevent panics.

nashley: Thanks for taking a look!
This _should_ be an infallible operation on `32` and `64` bit platforms. The only alternative I can think of is to wait on https://github.com/rust-lang/rust/issues/70460, but if you know of another way, I'd love to hear it.

dirvine: Thanks for the PR, nice one. Where the type is a result then `?` is OK here. 
Where we see a result we always use it or pass an error back (or resolve the error).

nashley: I added in a runtime check as well. Do we want to keep both the compile-time check and the (theoretically unreachable) runtime check?

src/wire_msg.rs
```diff
@@ -41,7 +42,16 @@ impl WireMsg {
...
         log::debug!("reading data of length: {}", msg_header.data_len());
-        let mut data: Vec<u8> = vec![0; msg_header.data_len()];
+        // https://github.com/rust-lang/rust/issues/70460 for work on a cleaner alternative:
+        #[cfg(not(any(target_pointer_width = "32", target_pointer_width = "64")))]
+        {
+            compile_error!("You need an architecture capable of addressing 32-bit pointers");
+        }
+        let data_length = match usize::try_from(msg_header.data_len()) {
+            Ok(s) => s,
+            Err(_) => return Err(Error::MaxLengthExceeded(usize::MAX)), // this should never happen due to preceding compilation error
```
nashley, line 52: Just realized that this new use (where it gives the maximum allowed length) as part of `MaxLengthExceeded` conflicts with the existing use (where it gives the specified length).

Which do we prefer? If the latter, this error needs to change types to `u32` or `u64` or something to be portable, but I believe that would be a breaking change?
I can also create a new dummy error type called "unreachable" or something and indicate that the user should file a bug report.

nashley: I think the only way to not change behavior here is to make a new error type.

loziniak:
I'd love this to be merged.

lionel1704:
Thanks for the poke @loziniak

@nashley could you give this a quick rebase. We can get this in.

lionel-faber (approved):

lionel1704:
I've rebased this and updated the commit message mentioning `BREAKING CHANGE` so that the next auto-deployed release is bumped up a minor version. Thanks for this @nashley :tada:

PR #220, opened 2021-02-17
feat: allow override of bootstap cache

joshuef:
Allow passing in bootstrap contact override

(used in client when attempting to relocate eg)

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -157,6 +157,10 @@ impl QuicP2p {
     /// Once a connection with a peer succeeds, a `Connection` for such peer will be returned
     /// and all other connections will be dropped.
     ///
+    /// `override_cache_addresses` can be used to bootstrap aginst specific addresses outwith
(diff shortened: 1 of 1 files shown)


Yoga07 (dismissed):

src/api.rs
```diff
@@ -175,12 +175,13 @@ impl QuicP2p {
...
     ///     config.local_port = Some(3001);
     ///     let mut quic_p2p = QuicP2p::with_config(Some(config), &[peer_addr], true)?;
-    ///     let endpoint = quic_p2p.bootstrap().await?;
+    ///     let endpoint = quic_p2p.bootstrap(None).await?;
     ///     Ok(())
     /// }
     /// ```
     pub async fn bootstrap(
         &self,
+        override_target_addresses: Option<&Vec<SocketAddr>>,
```
bochaco, line 184: I don't think we need the `Option` just a slice `&[SocketAddr]`, it's not empty you override.

PR #230, opened 2021-02-20
refactor(utils): moving bootstrap-cache exclusive utilities out of common utils mod

bochaco:


dirvine (approved):
Excellent and much more professional than a utils/common thing. Nice one

PR #243, opened 2021-03-11
Rebuild pre-async qp2p examples to use the new API

### LINKED ISSUE #167: Provide example that can send between computers
# Problem
I have been trying to get a simple file send happening between two peers on a local network and have been unable to make a connection.
It works fine when both sender and receiver are on the same computer but when i use two computers it simply doesn't connect
I don't actually know if this  is a bug or me not understanding.
Both computers should have no firewall and are on the same network switch.
# Example
## Client
```rust
async fn client( addr:&String,filePath:&str) -> Result<(), Error> {

    let mut file= std::fs::File::open(filePath)?;
    
    let serveraddr=addr.parse().expect("failed parsing server ip");

    let mut config = Config::default();
    config.ip = Some(IpAddr::V4(Ipv4Addr::LOCALHOST));
    config.port=Some(0);

    let mut quic_p2p =qp2p::QuicP2p::with_config(Some(config), Default::default(), true)?;

    //connect to the server
    let peer2= quic_p2p.new_endpoint().expect("failed creating endpoint");
    println!(" CLIENT HOSTING AT: {}",peer2.our_endpoint().unwrap());
    println!("Client connecting to : {}",serveraddr);
    let mut connection= peer2.connect_to(&serveraddr).await?;
    //send files 
}
```
## Server
```rust
async fn server_recv_file(host_ip: &String) -> Result<(),Error> {
    println!("Hello, world!");
    
    let mut file=std::fs::File::create("./out")?;

    let con=qp2p::QuicP2p::new()?;
    
    let mut config = Config::default();
    let addr:Ipv4Addr=host_ip.parse().expect("could not parse input ip addr 'x.x.x.x'");
    config.ip = Some(addr.into());
    config.port=Some (5500);
    let mut quic_p2p = QuicP2p::with_config(Some(config.clone()), Default::default(), true)?;
    let server=quic_p2p.new_endpoint()?;
    println!("HOSTING SERVER AT: {}",server.our_endpoint().unwrap());
    //configure listener
    let mut listener=server.listen()?;
    let mut a=listener.next().await.unwrap();
    println!("recived connection");
    //receive `files`
}
```

lionel-faber:
Resolves #167


examples/chat.rs
```diff
@@ -0,0 +1,256 @@
+// Copyright 2021 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under the MIT license <LICENSE-MIT
+// http://opensource.org/licenses/MIT> or the Modified BSD license <LICENSE-BSD
+// https://opensource.org/licenses/BSD-3-Clause>, at your option. This file may not be copied,
+// modified, or distributed except according to those terms. Please review the Licences for the
+// specific language governing permissions and limitations relating to use of the SAFE Network
+// Software.
+
+//! Basic chat like example that demonstrates how to connect with peers and exchange data.
```
dan-da, line 10: I would add a note here that this example is P2P and does NOT use/require the bootstrap server example.

examples/bootstrap_node.rs
```diff
@@ -0,0 +1,100 @@
+// Copyright 2019 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under the MIT license <LICENSE-MIT
+// http://opensource.org/licenses/MIT> or the Modified BSD license <LICENSE-BSD
+// https://opensource.org/licenses/BSD-3-Clause>, at your option. This file may not be copied,
+// modified, or distributed except according to those terms. Please review the Licences for the
+// specific language governing permissions and limitations relating to use of the SAFE Network
+// Software.
+
+//! Bootstrap node which acts as a relay for other client nodes. It collects the info of multiple
```
dan-da, line 10: Perhaps add a note that this bootstrap example is intended to be used with client_node.rs example.

examples/common/mod.rs
```diff
@@ -0,0 +1,95 @@
...
+                if let Some(addr) = addr {
+                    Some(Event::ConnectedTo { addr })
+                } else {
+                    None
+                }
+            }
+        }
+    }
+
+    // #[allow(unused)]
```
dan-da, line 66: let's get rid of commented code.

dan-da (changes requested):
These are examples for people new to qp2p, so I would suggest that all structs and all functions be commented at to their purpose/function.

Other than that, looks good, and it's very helpful to have working examples, so it would be great to get this merged asap.

dan-da:
So I am trying out the chat example, but haven't got two peers to connect yet.  Maybe I'm doing something wrong?

**Peer A**
```
cargo run --example chat -- --local-ip 127.0.0.1 --local-port 12000
```

**Peer B**
```
cargo run --example chat -- --local-ip 127.0.0.1 --local-port 12001
>> addpeer 127.0.0.1:12000
>> listpeers
  0: client:127.0.0.1:12000
>> send 0 hello
Error: No connection to the dest peer
```

**Peer A**
```
>> listpeers
>>
>> ourinfo
Our info: 127.0.0.1:12000
```

If there's not a bug, please post working usage example.

dan-da:
I was able to get the chat example working, at least on localhost.

It was never actually attempting to connect.  Now it calls endpoint.connect_to() during addpeer command.

Here's the patch.

```
diff --git a/examples/chat.rs b/examples/chat.rs
index 1c4ba2e..bd0169d 100644
--- a/examples/chat.rs
+++ b/examples/chat.rs
@@ -35,10 +35,6 @@ impl PeerList {
         Self { peers: Vec::new() }
     }
 
-    fn insert_from_str(&mut self, input: &str) -> Result<()> {
-        parse_peer(input).map(|v| self.insert(v))
-    }
-
     fn insert(&mut self, peer: SocketAddr) {
         if !self.peers.contains(&peer) {
             self.peers.push(peer)
@@ -108,7 +104,12 @@ async fn main() -> Result<()> {
                         print_ourinfo(&endpoint);
                         Ok(())
                     }
-                    "addpeer" => peerlist.insert_from_str(&args.collect::<Vec<_>>().join(" ")),
+                    "addpeer" => {
+                        let addr = parse_peer(&args.collect::<Vec<_>>().join(" "))?;
+                        endpoint.connect_to(&addr).await?;
+                        peerlist.insert(addr);
+                        Ok(())
+                    }
                     "listpeers" => {
                         peerlist.list();
                         Ok(())
```

PR #250, opened 2021-04-01
chore(rename) Rename forward_port to use_igd

dirvine:


Yoga07 (dismissed):

src/endpoint.rs
```diff
@@ -231,7 +231,7 @@ impl Endpoint {
         let mut addr = None;
 
         #[cfg(feature = "no-igd")]
-        if self.qp2p_config.forward_port {
+        if self.qp2p_config.use_igd {
             warn!("Ignoring 'forward_port' flag from config since IGD has been disabled (feature 'no-igd' has been set)");
```
dan-da, line 235: should be "Ignoring 'use_igd' flag"

src/config.rs
```diff
@@ -45,7 +45,7 @@ pub struct Config {
     /// Specify if port forwarding via UPnP should be done or not. This can be set to false if the network
```
dan-da, line 45: Perhaps "Specify if port forwarding via UPnP Internet Device Gateway Protocol (IGD) should be done or not..."

PR #251, opened 2021-04-01
fix: check local addr is not loopback when no public is known

joshuef:


oetyng (dismissed):

oetyng (approved):

src/endpoint.rs
```diff
@@ -165,6 +169,9 @@ impl Endpoint {
             } else {
                 warn!("Public IP address not verified since bootstrap contacts are empty");
             }
+        } else if !local_addr.ip().is_loopback() {
+            info!("local addr is not loopback: {:?}", local_addr);
+            endpoint.public_addr = Some(local_addr);
```
lionel1704, line 174: This seems wrong. If we have a local IP passed here this will set the external IP = local IP and it will bite us later. I'll look at today's hurdles and see what can be done now :+1:

oetyng (changes requested):
See @lionel1704's comment.

PR #262, opened 2021-04-12
feat(endpoint): add functionality for re-boostrapping to the network using updated contacts

lionel-faber:


S-Coyle (dismissed):

src/api.rs
```diff
@@ -97,15 +97,11 @@ impl QuicP2p {
 
         let mut bootstrap_cache =
             BootstrapCache::new(cfg.hard_coded_contacts, custom_dirs.as_ref())?;
-        if use_bootstrap_cache {
```
bochaco, line 100: I think it'll be useful o have some log entry stating that bootstrap cache is being used and which are the nodes being used from the cache.

src/tests/common.rs
```diff
@@ -423,6 +423,7 @@ async fn many_messages() -> Result<()> {
 // bootstrap contacts later.
 #[tokio::test]
 async fn connection_attempts_to_bootstrap_contacts_should_succeed() -> Result<()> {
+    utils::init_logging();
```
Yoga07, line 426: Should this be done for other tests too or should be removed?

lionel1704: It is already done in most tests. This particular one was failing so I added it in for debugging. It can be added in when needed.

Yoga07 (approved):

PR #263, opened 2021-04-12
feat(api): add new api to check if endpoint is externally reachable

lionel-faber:


src/endpoint.rs
```diff
@@ -380,6 +380,32 @@ impl Endpoint {
         Ok(())
     }
 
+    /// Verify if an address is publicly rechable. This will attempt to create
```
maqi, line 383: reachable

src/tests/common.rs
```diff
@@ -444,3 +444,19 @@ async fn connection_attempts_to_bootstrap_contacts_should_succeed() -> Result<()
...
+    let (ep2, _, _, _) = qp2p.new_endpoint().await?;
+
+    if let Ok(()) = ep1.is_reachable(&"127.0.0.1:12345".parse()?).await {
+        anyhow!("Unexpected success");
+    };
+    let reachable_addr = ep2.socket_addr();
+    ep1.is_reachable(&reachable_addr).await?;
+    Ok(())    
+
+}
```
maqi, line 462: missing a line break?

maqi (approved):

PR #273, opened 2021-06-08
multiple_connections_with_many_concurrent_messages test

maqi:


src/tests/common.rs
```diff
@@ -357,6 +357,73 @@ async fn multiple_concurrent_connects_to_the_same_peer() -> Result<()> {
...
 
+#[tokio::test]
+async fn multiple_connections_with_many_concurrent_messages() -> Result<()> {
+    use futures::future;
+
+    utils::init_logging();
+
+    let num_senders: usize = 10;
+    let num_messages_each: usize = 10;
+    let num_messages_total: usize = 100;
```
dirvine, line 368: Cold we make this 100 and 1000?

src/tests/common.rs
```diff
@@ -357,6 +357,73 @@ async fn multiple_concurrent_connects_to_the_same_peer() -> Result<()> {
...
+
+    let qp2p = new_qp2p()?;
+    let (recv_endpoint, _, mut recv_incoming_messages, _) = qp2p.new_endpoint().await?;
+    let recv_addr = recv_endpoint.socket_addr();
+
+    let test_msgs: Vec<_> = (0..num_messages_each).map(|_| random_msg()).collect();
+    let sending_msgs = test_msgs.clone();
```
dirvine, line 375: What size are these?

src/tests/common.rs
```diff
@@ -357,6 +357,73 @@ async fn multiple_concurrent_connects_to_the_same_peer() -> Result<()> {
...
+    let mut tasks = Vec::new();
+
+    // Receiver
+    tasks.push(tokio::spawn({
+        async move {
+            let mut num_received = 0;
+
+            while let Some((src, msg)) = recv_incoming_messages.next().await {
+                log::info!("received from {:?} with message size {}", src, msg.len());
+                assert_eq!(msg.len(), test_msgs[0].len());
```
dirvine, line 386: We were supposed to get the msg, hash it 5 times and send back the result.

src/tests/common.rs
maqi: 1KBytes

joshuef:
ideally here we want to have messages back to the senders confirmed.

probably with arbitrary delay in between.

Ideally we want to test that w/ a connection timeout from some sender, (so after 30s from disconnect), that another sender can still receive messages.

joshuef (approved):

PR #279, opened 2021-06-16
Add test to exchange multiple messages using quinn API

lionel-faber:
- test(quinn): replicate test that sends exchanges multiple messages using the quinn API directly

### DIFF
--- a/src/tests/quinn.rs
+++ b/src/tests/quinn.rs
@@ -12,7 +12,7 @@ use std::collections::BTreeSet;
 use tokio::sync::mpsc::{channel, Receiver, Sender};
 use tokio::sync::mpsc::{unbounded_channel, UnboundedReceiver, UnboundedSender};
 use tokio::sync::RwLock;
-use tracing::{error, trace, warn};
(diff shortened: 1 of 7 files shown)


src/tests/quinn.rs
```diff
@@ -0,0 +1,299 @@
...
+                let bytes = recv.read_to_end(1024).await?;
+                message_tx.send((peer_addr, Bytes::from(bytes))).await;
+            }
+        }
+    }
+    Ok(())
+}
+
+#[tokio::test(flavor = "multi_thread")]
+#[ignore = "Will run into OOM on CI"]
```
joshuef, line 201: How much mem are we using here? 

would this pass CI w/ less hashing?

Seems worrying that this would OOM on CI. (or does it just timeout?)

lionel1704: We can get this passing by reducing the number of sender and/or messages. 

It OOMs, I've seen `failed to allocate 123456 bytes` on CI with this test

joshuef (approved):

PR #282, opened 2021-06-28
fix(endpoint): close quinn endpoint when the Endpoint object is dropped

lionel-faber:


joshuef (dismissed):

Yoga07 (approved):

lionel1704:
This fix is incorrect. `qp2p::Endpoint` implements clone so closing the `quinn::Endpoint` on drop() will terminate the endpoint regardless of other instances of `qp2p::Endpoint`.

PR #284, opened 2021-07-01
feat(error): add error variant when creating port binding fails

lionel-faber:


src/igd.rs
```diff
@@ -27,6 +32,9 @@ pub async fn forward_port(local_addr: SocketAddr, lease_duration: u32) -> Result
 
             loop {
                 let _ = timer.tick().await;
+                if let Err(TryRecvError::Empty) = termination_rx.try_recv() {
```
connec, line 35: Am I missing something, or is this the wrong way around? E.g. currently it will `break` if the channel is empty – meaning `close` was not called.

Should it be `!=` or `Ok(_) =` instead? E.g. `break` if the channel *isn't* empty – meaning `close` was called (or some other error).

connec: I guess probably `!=` since other errors should (presumably) break the loop:

```rust
if termination_rx.try_recv() != Err(TryRecvError::Empty) {
    // endpoint was closed or dropped
    break;
}
```

dirvine (approved):

PR #289, opened 2021-07-12
Test for errors during large message send/receipt + retry on send fail

joshuef:
Close issue https://github.com/maidsafe/safe_network/issues/69

- adds retry attempts on new connections
- adds retry atetmpts on sending messages
- adds test for sending many 1mb messages

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 Cargo.lock
 *.*~
 .idea
+*.log
--- a/src/connection_deduplicator.rs
+++ b/src/connection_deduplicator.rs
@@ -73,7 +73,7 @@ impl ConnectionDeduplicator {
 pub(crate) enum Error {
     #[error("Connect error")]
     Connect(#[from] quinn::ConnectError),
-    #[error("Connection error")]
+    #[error("Quinn Connection error during deduplication")]
     Connection(#[from] quinn::ConnectionError),
 }
 
@@ -81,7 +81,7 @@ impl From<Error> for crate::error::Error {
     fn from(src: Error) -> Self {
         match src {
             Error::Connect(source) => Self::Connect(source),
-            Error::Connection(source) => Self::Connection(source),
+            Error::Connection(_) => Self::QuinnConnectionClosed,
         }
     }
 }
--- a/src/connection_deduplicator.rs
+++ b/src/connection_deduplicator.rs
@@ -81,7 +81,7 @@ impl From<Error> for crate::error::Error {
     fn from(src: Error) -> Self {
         match src {
             Error::Connect(source) => Self::Connect(source),
-            Error::Connection(_) => Self::QuinnConnectionClosed,
+            Error::Connection(source) => Self::Connection(source),
         }
     }
 }
--- a/src/connections.rs
+++ b/src/connections.rs
@@ -19,7 +19,7 @@ use futures::{future, stream::StreamExt};
 use std::net::SocketAddr;
 use tokio::sync::mpsc::{Receiver, Sender};
 use tokio::time::{timeout, Duration};
-use tracing::{debug, error, trace, warn};
+use tracing::{trace, warn};
 
 /// Connection instance to a node which can be used to send messages to it
 #[derive(Clone)]
@@ -28,15 +28,13 @@ pub(crate) struct Connection {
     remover: ConnectionRemover,
 }
 
-pub type DisconnectSender = Sender<(SocketAddr, Result<()>)>;
 /// Disconnection events, and the result that led to disconnection.
-pub struct DisconnectionEvents(pub Receiver<(SocketAddr, Result<()>)>);
+pub struct DisconnectionEvents(pub Receiver<SocketAddr>);
 
 /// Disconnection
-
 impl DisconnectionEvents {
     /// Blocks until there is a disconnection event and returns the address of the disconnected peer
-    pub async fn next(&mut self) -> Option<(SocketAddr, Result<()>)> {
+    pub async fn next(&mut self) -> Option<SocketAddr> {
         self.0.recv().await
     }
 }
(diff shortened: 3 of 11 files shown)


src/connection_deduplicator.rs
```diff
@@ -73,15 +73,21 @@ impl ConnectionDeduplicator {
...
 
 impl From<Error> for crate::error::Error {
     fn from(src: Error) -> Self {
         match src {
             Error::Connect(source) => Self::Connect(source),
-            Error::Connection(source) => Self::Connection(source),
+            Error::Connection(source) => match source {
+                quinn::ConnectionError::ConnectionClosed(_) => Self::QuinnConnectionClosed,
+                err => {
+                    tracing::warn!("** ???? ?the conn err: {:?}", err);
```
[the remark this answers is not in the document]
joshuef: sorry, this shouldn't have been marked as ready for review yet!

src/endpoint.rs
```diff
@@ -41,7 +41,10 @@ const PORT_FORWARD_TIMEOUT: u64 = 30;
 const ECHO_SERVICE_QUERY_TIMEOUT: u64 = 30;
 
 /// Standard size of our channel bounds
-const STANDARD_CHANNEL_SIZE: usize = 1000;
+const STANDARD_CHANNEL_SIZE: usize = 10000;
+
+/// Max number of attempts for connection retries
+const MAX_ATTEMPTS: usize = 10;
```
Yoga07, line 47: Would `10` be too much? Meaning, it can take some time to complete all 10 retries if there's the receiver is gone surely. Maybe a smaller number like 3 or 5?

src/connections.rs
```diff
@@ -41,9 +52,11 @@ impl Connection {
     /// Send message to peer using a uni-directional stream.
     pub async fn send_uni(&self, msg: Bytes) -> Result<()> {
         let mut send_stream = self.handle_error(self.quic_conn.open_uni().await).await?;
-        self.handle_error(send_msg(&mut send_stream, msg).await)
+        self.handle_error(send_msg(&mut send_stream, msg.clone()).await)
             .await?;
 
+        trace!("Message sent");
```
maqi, line 58: this log will not be that useful for production code. 
maybe remove it or make it log some more useful info?

src/endpoint.rs
```diff
@@ -500,7 +517,18 @@ impl Endpoint {
...
         self.connect_to(dest).await?;
-        self.try_send_message(msg, dest).await
+
+        let mut attempts: usize = 0;
+        let mut res = self.try_send_message(msg.clone(), dest).await;
+
+        while attempts < MAX_ATTEMPTS && res.is_err() {
+            trace!("send attempt # {:?}", attempts);
+            attempts += 1;
+            sleep(Duration::from_millis(500)).await;
```
Yoga07, line 527: Could we have this configurable? `RETRY_INTERVAL`?

src/connections.rs
```diff
@@ -41,9 +52,11 @@ impl Connection {
     /// Send message to peer using a uni-directional stream.
     pub async fn send_uni(&self, msg: Bytes) -> Result<()> {
         let mut send_stream = self.handle_error(self.quic_conn.open_uni().await).await?;
-        self.handle_error(send_msg(&mut send_stream, msg).await)
+        self.handle_error(send_msg(&mut send_stream, msg.clone()).await)
```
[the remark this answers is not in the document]
joshuef: Bytes clone is cheap, so not that much of an issue. But aye looks like its no longer needed here

.gitignore
```diff
@@ -3,3 +3,4 @@
 Cargo.lock
 *.*~
 .idea
+*.log
```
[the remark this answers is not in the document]
joshuef: how do you mean?

Yoga07 (dismissed):

src/connections.rs
joshuef: removed

.gitignore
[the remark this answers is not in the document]
maqi: I mean it needs an extra line break at the end of file.
otherwise it will show this red icon.

src/endpoint.rs
```diff
@@ -500,7 +517,18 @@ impl Endpoint {
             return Ok(());
         }
         self.connect_to(dest).await?;
-        self.try_send_message(msg, dest).await
+
+        let mut attempts: usize = 0;
+        let mut res = self.try_send_message(msg.clone(), dest).await;
+
+        while attempts < MAX_ATTEMPTS && res.is_err() {
```
maqi, line 524: we will try resends for 10 times, then on disconnection, try re-connect for 10 times as well ?
and each takes an interval of 0.5s, which makes a total of 50s to detect a really failed connection ?

src/connections.rs
```diff
@@ -193,15 +201,22 @@ pub(super) fn listen_for_incoming_messages(
 ) {
     let src = *remover.remote_addr();
     let _ = tokio::spawn(async move {
-        debug!("qp2p another incoming listerner spawned");
-        let _ = future::join(
+        match future::try_join(
```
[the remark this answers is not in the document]
joshuef: Yeh the other future will be cancelled as I understand.

src/connections.rs
```diff
@@ -288,19 +311,24 @@ async fn read_on_bi_streams(
...
+                        break;
+                    }
                     Err(err) => {
-                        error!(
-                            "Failed reading from a bi-stream for peer {:?} with error: {:?}",
+                        warn!(
+                            "Failed reading from a bi-stream for peer {:?} with: {:?}",
                             peer_addr, err
                         );
                         break;
```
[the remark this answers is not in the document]
joshuef: As I can figure, it's a question of "can this connection still be used". If so, we do not error.

src/endpoint.rs
joshuef: Could be. (worth nothing, this has already been updated to `MAX_ATTEMPTS=5` ). 

What's an acceptable time here? How many tries do we want? 

50s doesn't seem too bad. _to be sure_, vs just a 30s timeout eg

src/connections.rs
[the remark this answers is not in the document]
joshuef: So we were breaking before here, so that's not a change. 

As I understand, here we've received the bytes and it's the processing that fails (some message fails), but the connection is still valid, so we don't want to err out and cause a disconnect.

src/endpoint.rs
maqi: mainly just confirming this setup is ok.

maqi (approved):

src/connections.rs
```diff
@@ -28,6 +28,17 @@ pub(crate) struct Connection {
     remover: ConnectionRemover,
 }
 
+/// Disconnection events, and the result that led to disconnection.
```
lionel1704, line 31: Outdated comment?

PR #295, opened 2021-07-27
feat(id): keep a track of connections using a connection ID within qp2p

lionel-faber:


src/endpoint.rs
```diff
@@ -476,8 +478,8 @@ impl Endpoint {
     }
 
     /// Get an existing connection for the peer address.
-    pub(crate) async fn get_connection(&self, peer_addr: &SocketAddr) -> Option<Connection> {
-        if let Some((conn, guard)) = self.connection_pool.get(peer_addr).await {
+    pub(crate) async fn get_connection(&self, peer_addr: &SocketAddr) -> Option<Connection<I>> {
```
joshuef, line 481: we probably need `get_connection_by_id` too no? 

In Safe we'd want to avoid any referencing via `SockerAddr`

src/connection_pool.rs
```diff
@@ -14,69 +14,90 @@ use tokio::sync::RwLock;
...
         // Efficiently fetch the first entry whose key is equal to `key` and check if it exists
         store
-            .map
+            .key_map
             .range(Key::min(*addr)..=Key::max(*addr))
             .next()
             .is_some()
     }
 
+    #[allow(unused)]
```
connec, line 61: Nitpick: This should be unnecessary for a `pub fn`.

src/connection_pool.rs
```diff
@@ -14,69 +14,90 @@ use tokio::sync::RwLock;
 // Pool for keeping open connections. Pooled connections are associated with a `ConnectionRemover`
 // which can be used to remove them from the pool.
 #[derive(Clone)]
-pub(crate) struct ConnectionPool {
-    store: Arc<RwLock<Store>>,
+pub(crate) struct ConnectionPool<T: Id> {
```
connec, line 17: Nitpick: we probably should be consistent with the type parameter name. We're mostly using `I: Id`, which I think is a bit better naming.

connec (commented):
I've left some non-functional nitpicks.

Another thing I wondered would be whether we should give `I` a default type? Perhaps `impl Id for u64` and `I = u64` (maybe we could even remove `Key`?).

May not be worth it though, since this would already be a breaking change with the method renames.

src/connection_pool.rs
```diff
@@ -14,69 +14,90 @@ use tokio::sync::RwLock;
...
         store
-            .map
+            .key_map
             .range(Key::min(*addr)..=Key::max(*addr))
             .next()
             .is_some()
     }
 
+    #[allow(unused)]
+    pub async fn has_id(&self, id: &T) -> bool {
```
bochaco, line 62: perhaps we should rename it to `has_conn_id` ? and also perhaps have the methods like this one to expect an arg of type `&dyn ConnId` rather than having a generic on the whole structs?

src/connection_pool.rs
```diff
@@ -103,11 +126,21 @@ impl ConnectionRemover {
...
-    map: BTreeMap<Key, quinn::Connection>,
+struct Store<T: Id> {
+    id_map: BTreeMap<T, (quinn::Connection, Key)>,
+    key_map: BTreeMap<Key, (quinn::Connection, T)>,
     id_gen: IdGen,
 }
 
+/// Unique key identifying a connection. Two connections will always have distict keys even if they
+/// have the same socket address.
+pub trait Id:
```
bochaco, line 137: How about naming it `ConnId` since it's what it is?

joshuef (approved):

PR #296, opened 2021-07-28
feat: retry connection attempts for new connections too

joshuef:


src/endpoint.rs
```diff
@@ -351,18 +349,13 @@ impl Endpoint {
...
-        let mut connecting = self.attempt_connection(node_addr).await;
-
-        while connecting.is_err() && attempts < MAX_ATTEMPTS {
-            sleep(Duration::from_millis(self.qp2p_config.retry_interval)).await;
-            connecting = self.attempt_connection(node_addr).await;
-            attempts += 1;
-        }
-
-        let final_conn = connecting?;
+        // self.attempt_connection(node_addr)try_with_backoff( || self.attempt_connection(node_addr)).await?;
```
connec, line 357: The commented out stuff should go I reckon. History is always available in git.

src/endpoint.rs
```diff
@@ -351,18 +349,13 @@ impl Endpoint {
             Some(Err(error)) => return Err(error.into()),
             None => {}
         }
+        let final_conn = retry(ExponentialBackoff::default(), || async {
```
connec, line 352: We could put this and the `create_new_connection` retry into `attempt_connection` instead, since it's called from both (and only by those). Since it's just changing what was there before though that could be a future thing.

src/endpoint.rs
```diff
@@ -518,15 +510,20 @@ impl Endpoint {
...
+        })
+        .await;
 
-        while attempts < MAX_ATTEMPTS && res.is_err() {
-            trace!("send attempt # {:?}", attempts);
-            attempts += 1;
-            sleep(Duration::from_millis(self.qp2p_config.retry_interval)).await;
-            res = self.try_send_message(msg.clone(), dest).await;
-        }
+        // let mut attempts: usize = 0;
```
connec, line 518: More commented out stuff to kill 💀

src/error.rs
```diff
@@ -18,6 +18,9 @@ pub type Result<T, E = Error> = std::result::Result<T, E>;
 #[derive(Debug, Error)]
 #[non_exhaustive]
 pub enum Error {
+    /// Coudl not connect even after retries
+    #[error("Connection could not be established after {0} attempts")]
+    ConnectionFailed(u32),
```
connec, line 23: I guess this didn't end up being used? I don't see it in the diff.

connec (changes requested):

connec (dismissed):
LGTM!

connec (approved):

PR #305, opened 2021-08-11
feat(priority): allow setting priority of a stream

oetyng:
Sending uni directional messages can now be done with a priority.
The initial value for any stream is `0`, and an `i32` is used to go above or
 below that value, while keeping an intuitive default.
We also expose this API on the public `SendStream` type, as it allows
downstream consumers more freedom in the usage of the feat.


src/error.rs
```diff
@@ -50,6 +50,9 @@ pub enum Error {
     /// Failed to create a new endpoint.
     #[error("Creating endpoint")]
     Endpoint(#[from] quinn::EndpointError),
+    /// Failed to set/get priority of stream.
+    #[error("Unknown stream, cannot set/get priority.")]
+    Priority,
```
lionel1704, line 55: Will renaming this to `UnknownStream` make more sense? Since that's the only possible error in this scenario

lionel-faber (approved):

PR #325, opened 2021-09-13
feat(retries): extend send msg api with retry cfg

oetyng:
- This allows caller to apply a per-msg retry policy.
BREAKING CHANGE: Removal of a public field plus `Eq` and `PartialEq` derivations on `Config`.


src/config.rs
```diff
@@ -147,7 +208,50 @@ pub(crate) struct InternalConfig {
...
+}
+
+impl Default for RetryConfig {
+    fn default() -> Self {
+        Self {
+            initial_retry_interval: Duration::from_millis(200),
+            max_retry_interval: Duration::from_secs(10),
+            retry_delay_multiplier: 1.5,
+            retry_delay_rand_factor: 0.3,
+            retrying_max_elapsed_time: Duration::from_secs(30),
```
grumbach, line 252:
```suggestion
            retrying_max_elapsed_time: DEFAULT_RETRYING_MAX_ELAPSED_TIME,
```
why not use these DEFAULTs?

src/config.rs
```diff
@@ -147,7 +208,50 @@ pub(crate) struct InternalConfig {
...
+    /// Retrying continues until this time has elapsed.
+    /// The number of retries before that happens, will be decided by the other retry config options.
+    pub retrying_max_elapsed_time: Duration,
+}
+
+impl Default for RetryConfig {
+    fn default() -> Self {
+        Self {
+            initial_retry_interval: Duration::from_millis(200),
+            max_retry_interval: Duration::from_secs(10),
```
grumbach, line 249:
```suggestion
            initial_retry_interval: DEFAULT_INITIAL_RETRY_INTERVAL,
            max_retry_interval: DEFAULT_MAX_RETRY_INTERVAL,
```
why are those values different?

grumbach (commented):
Couple of (probably naive) questions
- Why not use the DEFAULT macros for`Default::default` Trait?
- All the calls to `send_message` use `None` is the last arg really necessary?

joshuef (changes requested):
We need a `BREAKING CHANGE:` commit w/ the `Eq/PartialEq` being dropped

joshuef (dismissed):

src/config.rs
```diff
@@ -120,15 +141,55 @@ pub struct Config {
...
-    /// How long to retry establishing connections and sending messages.
+    /// The initial retry interval.
+    ///
+    /// This is the first delay before a retry, for establishing connections and sending messages.
+    /// The subsequent delay will be decided by the `retry_delay_multiplier`.
+    ///
+    /// If unspecified, this will default to [`DEFAULT_INITIAL_RETRY_INTERVAL`].
+    #[serde(default)]
+    #[structopt(long, parse(try_from_str = parse_millis), value_name = "MILLIS")]
+    pub initial_retry_interval: Option<Duration>,
```
connec, line 152: We could go for `pub retry: Option<RetryConfig>`. It's not much worse for callers, since they can do:

```rust
Config {
    retry: RetryConfig {
        field_i_care_about: Duration::from_secs(123),
        ..Default::default()
    }
}
```

But it makes quite a lot of stuff simpler, deduplicates docs, etc.

connec (commented):
Small API suggestion.

src/config.rs
oetyng: Yes I thought of that as well, but then there was some structop thing and I just skipped it for now.
We can do a follow up with that.

Btw, ` ..Default::default()` does that insert the defaults of the struct, or the defaults of the types of the fields?

connec: My only hesitance is that follow up would mean more breaking change and API churn (e.g. introducing new APIs here, starting to use them, then replacing them all soon after). Perhaps we could remove `StructOpt` instead? It seems quite misplaced here, but I'm not sure what could be using it (`safe_network` at least [seems safe](https://github.com/maidsafe/safe_network/blob/2b29dbbcb6b9ef5c1a5b3758511dd23aa32a7c60/src/node/config_handler.rs#L129-L131)).

I guess is removing `StructOpt` looks non-trivial, we should deal with it as a follow-up, but if it's just a case of dropping it because we don't use it, I think that would be best.

oetyng: `Structop` dropped now.

src/config.rs
```diff
@@ -147,7 +152,51 @@ pub(crate) struct InternalConfig {
...
+    /// The randomization factor to use for creating a range around the retry interval.
+    ///
+    /// A randomization factor of 0.5 results in a random period ranging between 50% below and 50%
+    /// above the retry interval.
+    pub retry_delay_rand_factor: f64,
+    /// The maximum elapsed time after instantiating
+    ///
+    /// Retrying continues until this time has elapsed.
+    /// The number of retries before that happens, will be decided by the other retry config options.
+    pub retrying_max_elapsed_time: Duration,
```
connec, line 187: Another reason I ended up with the `min_retry_duration` name is that [the implementation](https://docs.rs/backoff/0.3.0/src/backoff/exponential.rs.html#111-124) uses it like a minimum:

```rust
match self.max_elapsed_time {
    Some(v) if self.get_elapsed_time() > v => None,
    _ => {
        let random = rand::random::<f64>();
        let randomized_interval = Self::get_random_value_from_interval(
            self.randomization_factor,
            random,
            self.current_interval,
        );
        self.current_interval = self.increment_current_interval();
        Some(randomized_interval)
    }
}
```

E.g., retry until more than `max_elapsed_time` has passed, so it's really the minimum amount of time.

connec (approved):

src/config.rs
connec: I think it could be argued that's a bug upstream.

oetyng: That's because it would be very bad to put

`if self.get_elapsed_time() == v`

..since you would then probably never hit that.

So, such cases always use `>=` or `>`, even though it's conceptually a max time that is modeled.

`Max` elapsed time is more accurate because there is not going to be any more time spent retrying than that.
If the retried call succeeds though, it can be a lot shorter than that, so `Min` is not very accurate.

I think they have it right there.

connec: I think the implementation is fine, it's just the name sets the wrong expectation for me. "Max elapsed time" makes me think it will not retry for longer than that, when in reality it will retry for _at least_ that long, but in fact strictly longer.

PR #350, opened 2021-10-27
Disable keep-alives by default

connec:
- 10fa647 **fix: more carefully handle 'benign' connection loss**

  This changes the behaviour of the `ConnectionIncoming` pseudo-stream to
  signal end-of-stream when the connection closes for a 'benign' reason,
  rather than returning an error. This makes it much easier to implement
  message polling for callers, without them having to apply their own
  filtering.
  
  Since we're doing this detection in a couple of places, it made sense to
  move the check into `ConnectionError` itself.

- e511c59 **test: rewrite large message test with more concurrency**

  The old `IncomingMessages` stream would multiplex over all incoming
  connections. When we removed this, we inadvertently made this test
  process each connection fully, and sequentially, which leads to hitting
  various timeouts (both the timeouts in the test, and idle
  timeouts on the connection) more often.
  
  Concurrency has been restored with
  `TryStreamExt::try_for_each_concurrent`, which is used to process all
  incoming connections concurrently once again.

- f520fcf **refactor!: disable keep-alives by default**

  This changes the `keep_alive_interval` set for `Config::default()` to
  `None`, disabling keep-alives. This is thought to be a more sane default
  for a library, since keep-alives may otherwise cause connections to stay
  open forever, leaking resources. It remains easy to opt-in to
  keep-alives by simply setting `Some(duration)`.
  
  BREAKING CHANGE: The `qp2p::config::DEFAULT_KEEP_ALIVE_INTERVAL`
  constant has been removed, and the default value for
  `keep_alive_interval` set for `Config::default()` is now `None`, meaning
  keep-alives are disabled by default.

- 8186416 **feat: add `Connection::id` to get a stable ID for a connection**

  This just proxies the underlying `quinn::Connection::stable_id`. This is
  actually implemented in `quinn` as the address in memory of the
  underlying shared connection reference, so although it is stable for the
  lifetime of a connection it is not comparable across peers. It's still
  useful for tracking connections within a single peer.

- 79986c8 **chore: add more logging when handling echo/verify requests**

  Since these processes are transparent for callers, its easy to forget
  that they exist at all. This can lead to some confusion when trying to
  trace connection activity, since they might look spurious. Some
  additional logging here should help to clarify what's the craic.

- fe7450a **test: increase timeout for some tests**

  We want tests to fail, rather than hanging forever, in situation where
  we expect some incoming connection/message. Sadly, setting this timeout
  is a bit arbitrary, and depending on the performance of the system
  running the tests they could fire spuriously. That seems to be what
  we're seeing in CI, so we're bumping the timeout to compensate.
  
  Note that this will make some tests take longer, since the timeout is
  also used to check that something we don't expect to happen, doesn't
  happen.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -18,14 +18,6 @@ use std::{future::Future, net::IpAddr, sync::Arc, time::Duration};
 /// see no conversation between them.
 pub const DEFAULT_IDLE_TIMEOUT: Duration = Duration::from_secs(60);
 
-/// Default for [`Config::keep_alive_interval`] (20 seconds).
-pub const DEFAULT_KEEP_ALIVE_INTERVAL: Duration = Duration::from_secs(20);
-
-// serde needs a function when specifying a path for `default`
-fn default_keep_alive_interval() -> Option<Duration> {
-    Some(DEFAULT_KEEP_ALIVE_INTERVAL)
-}
-
 /// Default for [`Config::upnp_lease_duration`] (2 minutes).
 pub const DEFAULT_UPNP_LEASE_DURATION: Duration = Duration::from_secs(120);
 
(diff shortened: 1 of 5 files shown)


joshuef (dismissed):

Yoga07 (dismissed):

lionel-faber (commented):
Just one comment

lionel-faber (approved):

PR #358, opened 2021-11-12
Better retries

connec:
- bf2c26e **refactor!: rename `Config::retry_config` to `default_retry_config`**

  The plan is to introduce additional APIs to allow retries to be
  configured in all places that we perform retries. As such, the
  configured retry config will always behave as a default, and changing
  the name makes this clear.

- 027dc51 **refactor: improve retry APIs**

  This is an attempt to allow callers to maintain a 'retry context'
  throughout their own code and calls to `qp2p`. For example, it may be
  desired to retry an operation for 60s, regardless of where the failure
  occurs. Without a change such as this, callers would have to wrap their
  entire operation in a top level retry, and it becomes difficult to
  reason about where retries may occur and for how long.
  
  The APIs introduced here allow callers to supply an implementation of
  `backoff::backoff::Backoff`, which will be used to drive retries. That
  alone is insufficient to allow a 'retry context' to be maintained, but
  the included `qp2p::backoff::BackoffRef` utility struct allows a mutable
  reference to a `backoff::backoff::Backoff` to be passed.

### DIFF
--- a/src/backoff.rs
+++ b/src/backoff.rs
@@ -1,0 +1,25 @@
+//! Utilities for working with retries driven by the [`backoff`](::backoff) crate.
+
+use backoff::backoff::Backoff;
+use std::time::Duration;
+
+/// A wrapper for an `&mut impl `[`Backoff`].
+///
+/// It's useful to be able to supply a reference to a [`Backoff`] in order to maintain the same
+/// retry 'context' across multiple operations. Sadly this isn't possible with the trait
+/// implementations in the `backoff` crate itself, since `Backoff` is not implement for mutable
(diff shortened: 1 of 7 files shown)


connec:
Drafting this, as it may not be necessary just now and it's proved challenging to improve the ergonomics (see [this post](https://users.rust-lang.org/t/incompatible-lifetimes-mutable-references-in-an-async-loop/67269) for what I was trying, and how it could perhaps be worked around).

connec:
I'm gonna close this. It was not necessary, and getting it working would probably be more complicated than it's worth.

PR #366, opened 2022-01-18
feat!: Moving to quinn 0.8

joshuef:


src/config.rs
```diff
@@ -27,6 +24,12 @@ pub const DEFAULT_UPNP_LEASE_DURATION: Duration = Duration::from_secs(120);
 /// gives 5-6 retries in ~30 s total retry time.
 pub const DEFAULT_INITIAL_RETRY_INTERVAL: Duration = Duration::from_millis(500);
 
+/// Default for [`Config::idle_timeout`] (1 minute).
+///
+/// This is based on average time in which routers would close the UDP mapping to the peer if they
+/// see no conversation between them.
+pub const DEFAULT_IDLE_TIMEOUT: Duration = Duration::from_secs(60);
```
dirvine, line 31: This seems long, generally, 20-30 secs is the best we can hope for so setting this at 18 or so would make more sense. Otherwise, we think connections are open (to us) and they are closed. Not much on Inet about timeout, but https://www.quora.com/How-does-UDP-hole-punching-work

src/config.rs
```diff
@@ -65,17 +77,17 @@ impl From<rcgen::RcgenError> for ConfigError {
     }
 }
 
-impl From<quinn::ParseError> for ConfigError {
-    fn from(error: quinn::ParseError) -> Self {
-        Self::CertificateGeneration(CertificateGenerationError(error.into()))
-    }
-}
+// impl From<quinn::ParseError> for ConfigError {
```
dirvine, line 80: Was this intentional?

joshuef: ah, needs removed. 

quinn config is now infallible

joshuef:
So the config tweaks have been tested against ` multiple_connections_` tests. Those really hammer qp2p when you add more senders or messages. You can now bump then a faiiiir bit higher than main and not get fails. (Or i can! :D ymmv, but it feels a lot more stable for me)

If you check the `qp2p-<hash>` process, you can see what's going on there.

Without these changes I was seeing fails w/ a few repeat tests. Now I've had many many runs and not had a fail yet.



from config.rs here:

```

        // This has a bearing on max memory: https://docs.rs/quinn/latest/quinn/struct.TransportConfig.html#method.max_concurrent_bidi_streams
        // here we/ve lowered it from 1250000
        // eg this reduced max mem for: multiple_connections_with_many_larger_concurrent_messages
        // from ~1gb/22s to ~300mb/28s to complete (YMMV)
        let _ = config.stream_receive_window(VarInt::from_u32(500));

        // defaults here are 100, we've lowered this here to prevent hammering of nodes...
        // a lower count appears to reduce the chance of dropped connections quite significantly
        let _ = config.max_concurrent_bidi_streams(VarInt::from_u32(10));
        let _ = config.max_concurrent_uni_streams(VarInt::from_u32(10));
```

The uni stream limits seem pretty significant for me.

dirvine:
Brilliant news @joshuef Great bit of detective work there.

dirvine (dismissed):
LGTM, very good

src/config.rs
```diff
@@ -223,14 +221,41 @@ pub(crate) struct InternalConfig {
 
 impl InternalConfig {
     pub(crate) fn try_from_config(config: Config) -> Result<Self> {
-        let idle_timeout = config.idle_timeout.unwrap_or(DEFAULT_IDLE_TIMEOUT);
+        let default_idle_timeout: IdleTimeout = IdleTimeout::try_from(DEFAULT_IDLE_TIMEOUT)?; // 60s
```
maqi, line 224: DEFAULT_IDLE_TIMEOUT is 18s ?

src/error.rs
```diff
@@ -155,23 +147,16 @@ impl ConnectionError {
...
-                error_code: 0,
-                reason,
-            }) if reason.is_empty() => true,
-
-            // Any other error is classified as not benign, so should always be propagated.
-            _ => false,
-        }
+            Self::Closed(Close::Local) |
+            // A graceful close from the peer, with the default code 0
+            Self::Closed(Close::Application { error_code: 0, .. })
```
maqi, line 158: previously only `reason.is_empty()`  will be classified as benigh for Close::Application . 
now it is for as long as error_code is 0 ,
is it expected?

src/error.rs
```diff
@@ -187,7 +172,7 @@ impl From<quinn::ConnectError> for ConnectionError {
                 // error.
                 Self::InternalConfigError(InternalConfigError(error))
             }
-            quinn::ConnectError::Config(_) => {
+            _ => {
```
maqi, line 175: i'd prefer  an explicit entry instead of a wild card

src/tests/common.rs
```diff
@@ -712,7 +687,7 @@ async fn reachability() -> Result<()> {
 }
 
 #[tokio::test(flavor = "multi_thread")]
-#[traced_test]
+// #[traced_test]
```
maqi, line 690: why comment out?

src/config.rs
joshuef: Yeh, that came from @dirvine as a ostensibly a little inside the 20s that a lot of routers might cut off at. I'll add a comment to reflect this.

src/tests/common.rs
joshuef: ah, can be removed.

joshuef:
Okay so this is looking stable now. Seems like main factor in that stability was adding a `keep_alive` to the test endpoints.

My prior [assertion](https://github.com/maidsafe/qp2p/pull/366#issuecomment-1016596160) were true only while `tracing_test` was enabled. It looks like this slowed down the test runs _significantly_, so we were holding much more in memory for much longer. 

Impact with this disabled was minimal so these tweaks have been removed. (though they remain interesting candidates for optimisations).

PR #373, opened 2022-02-14
fix(p2p_node): 'polo' reply processing

m1cr0man:
When trying to run the p2p_node example, I encountered a crash when polo is sent back to the peer:

```bash
---
Listening on: 127.0.0.1:38616
---


Received from 127.0.0.1:38091 --> b"marco"
Error: 
   0: Connection was lost when trying to send a message
   1: The connection was closed by application (error code: 0, reason: )

```

The client shows:

```bash
Sending to 127.0.0.1:38616 --> b"marco"

Done sending

---
Listening on: 127.0.0.1:38091
---

```

I was able to get it working with the following patch. I'm a beginner to this library,
so please feel free to recommend a different way to tackle this! I was aiming for
the simplest solution in this context.

### DIFF
--- a/examples/p2p_node.rs
+++ b/examples/p2p_node.rs
@@ -53,7 +53,10 @@ async fn main() -> Result<()> {
                 .expect("Invalid SocketAddr.  Use the form 127.0.0.1:1234");
             let msg = Bytes::from(MSG_MARCO);
             println!("Sending to {:?} --> {:?}\n", peer, msg);
-            node.connect_to(&peer).await?.0.send(msg.clone()).await?;
+            let (conn, mut incoming) = node.connect_to(&peer).await?;
+            conn.send(msg.clone()).await?;
+            let reply = incoming.next().await?.unwrap();
+            println!("Received from {:?} --> {:?}", peer, reply);
         }
 
         println!("Done sending");


joshuef:
Thanks for this @m1cr0man . I can see that this might fix your issue (which I have seen prev but thought fixed 🤔 )...

But I'm not sure it's fixing the underlying issue (as opposed to the test)...

[here we're looping over these connections](https://github.com/maidsafe/qp2p/pull/373/files#diff-a5d6d6aa989ac73d8fc9851d9cf73fecee4602ee65eef464584b26f2a8f27557R74) and this _should_ do the same thing as you're after here (if I'm, reading it correctly). 

So I think something else is up here and we're perhaps just obfuscating that fact.


@maqi what do you make of this?

m1cr0man:
My understanding was that this is some sort of 0/1RTT case and because the "incoming" value of connect_to contains some data it must be processed explicitly and has already been popped out of incoming_messages. Maybe I'm wrong though.

m1cr0man:
Ugh sorry did not mean to close.. Mobile browsers are hard! 😓

maqi:
@m1cr0man 

According to your change, the client shall not print out 
```
Done sending

---
Listening on: 127.0.0.1:38091
---
```
before print out `Received from 127.0.0.1: 38616`
as it shall waiting for the incoming msg due to your change of `let reply = incoming.next().await?.unwrap();`

I tried out your change and get a printout from client as:
```
Sending to 127.0.0.1:50028 --> b"marco"

Received from 127.0.0.1:50028 --> b"polo"
Done sending

---
Listening on: 127.0.0.1:61983
---
```
And no error from server.

could you have a double check whether server/client are in the correct/same building version?
thx.

maqi:
@joshuef @m1cr0man 
So first I tried an old build of the example and it works.
With our newly change, seems the connection get `closed` too quick.
That is after the client called `node.connect_to(&peer).await?.0.send(msg.clone()).await?` , if the resulted `connection/listener` not get hold, the connection will get closed immediately.
Which means on the server side, when it attempt to send back, the connection actually get closed.

this example code may need to be refactored to reflect such behaviour (if this is expected)

joshuef:
Just to update @m1cr0man. Looks like this is indeed surfacing some legacy behaviour. 

We used to have a connection pool which would keep this alive for the tests, but now unless we manually hold the conn, it's dropped.

For now (assuming CI is happy) we'll get this in and the test sured up 👍 . 

@maqi will follow that up with some refactors to address this outdated behaviour.

Thanks for getting this in! 🙇

m1cr0man:
It looks like CI is failing on an unrelated file, but do you want me to quickly fix that before merge? It looks trivial

maqi:
> It looks like CI is failing on an unrelated file, but do you want me to quickly fix that before merge? It looks trivial

will be appricated to make CI all green. 
thank you very much.

maqi:
@m1cr0man @joshuef 
raised PR https://github.com/maidsafe/qp2p/pull/377 which contains the work of this PR,
together with extra work on commenting and resolving CI failure.

m1cr0man:
Sorry I didn't realize you were quite this eager to merge 😅 just finished work. I'll close this PR then

joshuef:
No worries @m1cr0man ! Thanks again for the PR 🙇

PR #374, opened 2022-02-22
Add API to respond to bidirectional stream when getting next message from connection.

hansonkd:
Hello,

I was building bindings for qp2p for Elixir and came across a case where i needed to not only accept the next message from a bidirectional stream, but also reply to it. I couldn't figure out a way with the current API to do this without having some type of correlation prefix on the message.

This PR adds a new method `next_with_stream` to `ConnectionIncoming` in order for users to respond to a bidirectional stream.

If you can accomplish this without these changes in the current API or if you have suggestions, please let me know!

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -478,7 +478,7 @@ async fn listen_on_bi_streams(
                     Ok(Some(WireMsg::EndpointVerificationReq(addr))) => {
                         if let Err(error) = handle_endpoint_verification(
                             endpoint,
-                            &mut (*arc_mutex.lock().await).inner,
+                            &mut arc_mutex.lock().await.inner,
                             addr,
                         )
                         .await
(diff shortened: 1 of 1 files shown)


dirvine (approved):
A nice PR, thank you. This one makes a lot of sense.

Yoga07:
Hi @hansonkd ,

Thank you for your contribution. Adding a new API does make better sense than complicating using prefixes, meaning Users can call whichever API they want accordingly. It would be nice if you can get the failing rustfmt failure [here](https://github.com/maidsafe/qp2p/runs/5337763723?check_suite_focus=true) resolved and we'd be good for a merge. 

Cheers!

joshuef:
also @hansonkd we [follow conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) for commit messages to auto generate changelogs for release. So if you could reformat your messages to follow that it would be great 🙇 

(eg: `feat: add stream..` )

joshuef:
This has been superseded by https://github.com/maidsafe/qp2p/pull/382 where I've tidied up the commit history and addressed the clippy issues w/r/t type complexity

PR #386, opened 2022-07-20
feat: add rebind api to remap SocketAddr in use

joshuef:


src/endpoint.rs
```diff
@@ -229,6 +229,15 @@ impl Endpoint {
         self.local_addr
     }
 
+    /// Change Endpoint SocketAddr
+    pub fn rebind_to(&self, local_addr: SocketAddr) -> Result<(), EndpointError> {
```
maqi, line 233: may the `self.public_addr` and `self.local_addr` also need to be updated?

PR #389, opened 2022-08-29
feat: take three Bytes for each send

joshuef:
This more closely reflects msg header, destintation and payload
parts, and allows for greater flexibility and deduplication when
sending repeat headers/payloads to different peers


src/wire_msg.rs
```diff
@@ -189,11 +259,15 @@ impl MsgHeader {
 
     fn from_bytes(bytes: [u8; MSG_HEADER_LEN]) -> Self {
         let version = u16::from_be_bytes([bytes[0], bytes[1]]);
-        let data_len = u32::from_be_bytes([bytes[2], bytes[3], bytes[4], bytes[5]]);
-        let usr_msg_flag = bytes[6];
+        let user_header_len = u32::from_be_bytes([bytes[2], bytes[3], bytes[4], bytes[5]]);
```
maqi, line 262: My understanding that for `ECHO_SRVC_MSG_FLAG` msgs, the `header` & `dst` will be `Bytes::new`.
However, seems here we still trying to parse from a 4 bytes segment ?

src/wire_msg.rs
```diff
@@ -67,19 +68,38 @@ impl WireMsg {
...
         let msg_flag = msg_header.usr_msg_flag();
 
-        recv.read_exact(&mut data).await?;
+        // fill up our data vecs from the stream
+        // ECHO_SRVC_MSG_FLAG does not have header
+        if header_length == 0 {
+            recv.read_exact(&mut header_data).await?;
+        }
+        // ECHO_SRVC_MSG_FLAG does not have dst
+        if dst_length == 0 {
```
b-zee, line 87: Shouldn't this be `!= 0`?

src/wire_msg.rs
b-zee: @maqi The lengths of the `header` and `dst` will always be in the `Msgheader`? In that case, these will both be `0x0000_0000`.

src/wire_msg.rs
```diff
@@ -67,19 +68,38 @@ impl WireMsg {
...
         } else if msg_flag == USER_MSG_FLAG {
-            Ok(Some(WireMsg::UserMsg(From::from(data))))
+            Ok(Some(WireMsg::UserMsg((
+                Bytes::from(header_data),
+                Bytes::from(dst_data),
+                Bytes::from(payload_data),
+            ))))
         } else if msg_flag == ECHO_SRVC_MSG_FLAG {
-            Ok(Some(bincode::deserialize(&data)?))
+            Ok(Some(bincode::deserialize(&payload_data)?))
```
b-zee, line 102: Purely from reading this line: I don't understand what it deserialises into. I assume the data will be a `SocketAddr`, but what type does Rust deserialise into, an `EndpointEchoResp` or `EndpointVerificationReq`? (They both have `SocketAddr`.)

src/wire_msg.rs
```diff
@@ -140,31 +184,47 @@ impl fmt::Display for WireMsg {
 #[derive(Debug)]
```
b-zee, line 184: Maybe change the documentation above here?

src/wire_msg.rs
joshuef: Yup. and we clearly need some more tests around this too!

src/wire_msg.rs
joshuef: It deserializes into `Self`, which'll be one of those enum variations. (though not the `UserMsg`

b-zee: Oh, I see now I reviewed this line without seeing what you changed. Your change is irrelevant to my comment, so feel free to ignore...

maqi (dismissed):

PR #398, opened 2022-11-08
Remove retry config

joshuef:


src/connection.rs
```diff
@@ -1,8 +1,8 @@
 //! A message-oriented API wrapping the underlying QUIC library (`quinn`).
 
 use crate::{
-    config::{RetryConfig, SERVER_NAME},
-    error::{ConnectionError, RecvError, RpcError, SendError, StreamError},
+    config::SERVER_NAME,
+    error::{ConnectionError, RecvError, RpcError, SendError, SerializationError, StreamError},
```
b-zee, line 5: Seems that this `SerializationError` was removed by @bochaco, not sure why you introduced it again here.

PR #405, opened 2022-12-05
fix(StreamID): wrap quinn::StreamID to avoid misleading log

maqi:


src/connection.rs
```diff
@@ -186,6 +186,88 @@ impl fmt::Debug for Connection {
...
+            dir,
+            self.index()
+        )
+    }
+}
+
+impl StreamId {
+    /// Create a new StreamId
+    pub fn new(initiator: Side, dir: Dir, index: u64) -> Self {
+        StreamId(index << 2 | (dir as u64) << 1 | initiator as u64)
```
bochaco, line 247: Why don't we simply use the enums instead of doing this type of encoding of these values?

bochaco (commented):
I'm also thinking we could use the conn id to also be part of the stream id. E.g. I'm seeing some connection to a client is closed and later on I see we fail to send a msg to that client over a stream, but I have not way to tell if the stream that failed was over the connection that I saw being closed beforehand.

src/connection.rs
```diff
@@ -186,6 +186,88 @@ impl fmt::Debug for Connection {
...
+    /// The initiator of a connection
+    Initiator = 0,
+    /// The acceptor of a connection
+    Acceptor = 1,
+}
+
+/// Whether a stream communicates data in both directions or only from the initiator
+#[cfg_attr(feature = "arbitrary", derive(Arbitrary))]
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub enum Dir {
```
bochaco, line 202: There is no harm in using full names for better clarity, i.e. `Direction`

src/connection.rs
```diff
@@ -186,6 +186,88 @@ impl fmt::Debug for Connection {
     }
 }
 
+/// Whether an endpoint was the initiator of a connection
+#[cfg_attr(feature = "arbitrary", derive(Arbitrary))]
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub enum Side {
```
bochaco, line 192: Perhaps this shall be called `StreamInitiator` for better clarity?

src/connection.rs
```diff
@@ -186,6 +186,88 @@ impl fmt::Debug for Connection {
...
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub struct StreamId(#[doc(hidden)] pub u64);
+
+impl fmt::Display for StreamId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let initiator = match self.initiator() {
+            Side::Initiator => "initiator",
+            Side::Acceptor => "acceptor",
+        };
+        let dir = match self.dir() {
```
bochaco, line 230: You don't need this, or even the `Display` impl for `Dir`, you can just use the `Debug` version of it in L236, i.e. `{dir:?}`.

src/connection.rs
maqi: That was copied from quinn's design.
I think it's trying to encode the info into the u64

maqi:
> I'm also thinking we could use the conn id to also be part of the stream id. E.g. I'm seeing some connection to a client is closed and later on I see we fail to send a msg to that client over a stream, but I have not way to tell if the stream that failed was over the connection that I saw being closed beforehand.

That then will require the connection_id to be passed down when contruct the stream.
will impact the API and prefer to be done as a separate PR?

b-zee:
Is this purely for printing out a more accurate format from our perspective? Isn't it possible to simply do this?

```
struct StreamId(quinn::StreamId);
impl fmt::Display for StreamId {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        let initiator = match self.0.initiator() {
            quinn_proto::Side::Client => "initiator",
            quinn_proto::Side::Server => "acceptor",
        };
        let dir = match self.0.dir() {
            quinn_proto::Dir::Uni => "uni",
            quinn_proto::Dir::Bi => "bi",
        };
        write!(
            f,
            "{} {}directional stream {}",
            initiator,
            dir,
            self.0.index()
        )
    }
}
```

maqi:
> Is this purely for printing out a more accurate format from our perspective? Isn't it possible to simply do this?
> 
> ```rust
> struct StreamId(quinn::StreamId);
> impl fmt::Display for StreamId {
>     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
>         let initiator = match self.0.initiator() {
>             quinn_proto::Side::Client => "initiator",
>             quinn_proto::Side::Server => "acceptor",
>         };
>         let dir = match self.0.dir() {
>             quinn_proto::Dir::Uni => "uni",
>             quinn_proto::Dir::Bi => "bi",
>         };
>         write!(
>             f,
>             "{} {}directional stream {}",
>             initiator,
>             dir,
>             self.0.index()
>         )
>     }
> }
> ```

yes, it is. The current logging from quinn::Streaming is bit confusing as using `Slient/Server` to describe actually the `Initiator/Acceptor` of a connection

src/connection.rs
maqi: This is for a more concised log for the StreamId, which is different to the default Debug or Dispay impl of `enum Dir`

src/connection.rs
bochaco: We shouldn't need such encoding here I believe?

src/connection.rs
bochaco: I think you don't need them, let me put an example further below in the other discussion

bochaco:
> prefer to be done as a separate PR?

I'll do that shortly

bochaco:
> Then what about my code? Doesn't that fulfill the same purpose? 

I agree with that impl above, we shouldn't need any of the additional enums as they are just to generate that string we use for logging purposes

b-zee (approved):

PR #446, opened 2023-02-15
feat: limit async during read/write for greater consistency

joshuef:


src/wire_msg.rs
```diff
@@ -88,13 +96,23 @@ impl WireMsg {
...
 
-        // Send message bytes over QUIC
-        send_stream.write_all(msg_head).await?;
-        send_stream.write_all(msg_dst).await?;
-        send_stream.write_all(msg_payload).await?;
+        let start = Instant::now();
+        // Send the header bytes over QUIC
+        send_stream.write_all(&all_bytes).await?;
+        let duration = start.elapsed();
+        trace!("Writing bytes took: {duration:?}");
```
bochaco, line 115: let's log the number of bytes written as well

src/wire_msg.rs
```diff
@@ -88,13 +100,23 @@ impl WireMsg {
...
```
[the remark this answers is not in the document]
joshuef: just for debugging if you want it? Could be removed, but yeh for now perhaps it's okay.

src/connection.rs
```diff
@@ -11,7 +11,7 @@ use tokio::sync::mpsc::{Receiver, Sender};
 use tracing::{trace, warn};
 
 // TODO: this seems arbitrary - it may need tuned or made configurable.
-const INCOMING_MESSAGE_BUFFER_LEN: usize = 10_000;
+const INCOMING_MESSAGE_BUFFER_LEN: usize = 1;
```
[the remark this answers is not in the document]
joshuef: Hmmm, I think it's a Q of feedback, again, if we have a buffer we'll never feedback over the wire.

b-zee (dismissed):

src/connection.rs
```diff
@@ -160,12 +160,25 @@ fn listen_on_uni_streams(connection: quinn::Connection, tx: Sender<IncomingMsg>)
...
+
+                let msg = WireMsg::read_from_stream(recv).await;
 
                 // Send away the msg or error
-                let _ = tx.send(msg.map(|r| (r, None))).await;
+                reserved_sender.send(msg.map(|r| (r, None)));
             });
+
+            // now that work is off thread, and we loop, a tiny pause for CPU sanity
+            tokio::time::sleep(Duration::from_millis(50)).await;
```
bochaco, line 181: I don't think we need this, or we shouldn't, if buffer is full then it'll be awaiting when attempting to reserve a lot, otherwise there shouldn't be anything needing such a thing

src/wire_msg.rs
```diff
@@ -49,30 +66,21 @@ impl WireMsg {
...
-        let mut header_data = BytesMut::with_capacity(header_length);
-        let mut dst_data = BytesMut::with_capacity(dst_length);
-        let mut payload_data = BytesMut::with_capacity(payload_length);
-        // buffer capacity does not actually give us length, so this sets us up
-        header_data.resize(header_length, 0);
-        dst_data.resize(dst_length, 0);
-        payload_data.resize(payload_length, 0);
+        // Check we have all the data and we weren't cut short, otherwise
+        // the following would panic...
+        if !bytes.len() < (header_length + dst_length + payload_length) {
```
bochaco, line 71: Shouldn't we compare exact length? ...if not then cleaner if we use `>=` instead of negating `<` comparison?

bochaco: ...I'm not sure actually if the condition is correct, using `>=` doesn't sound like what we want to either....?.....

bochaco (approved):

